PR #3, opened 2017-07-28
CI: Upload conda package for Happi to DEV and TAG channels

hhslepicka:


teddyrendahl:
Looks awesome man. My only thought is should we change the Travis to install `happi` via conda instead of `setup.py` that we verify the conda install works as well for tags?

hhslepicka:
@teddyrendahl I will look into that...

PR #5, opened 2017-08-24
ENH: Support for JSON 

### LINKED ISSUE #2: Equivalance method returns tuple

teddyrendahl:
* Abstracted all calls to MongoDB so we can switch back and forth between whatever backend we want
* Added backend support for MongoDB and JSON
* Setup landscape.io
* Closes #2 that deals with device comparison

My major concern with the JSON backend is we essentially have to destroy and rewrite the file every edit. I used some file locking functions to make sure we can't write at the same time and corrupt the file, but it still feels kind of scary.

### DIFF
--- a/.landscape.yml
+++ b/.landscape.yml
@@ -1,0 +1,10 @@
+doc-warnings : true
+test-warnings : false
+strictness: medium
+max-line-length : 80
+autodetect: true
+ignore-paths:
+   - docs
+python-targets:
+    - 2
+    - 3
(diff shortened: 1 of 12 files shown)


ZLLentz:
This looks good to me. There might be some merit in storing each object in a separate json file instead of in a monolith, since that's a closer 1-to-1 mapping with the mongodb documents, but this looks solid and it will let us host an nfs-db for our tiny database.

teddyrendahl:
@ZLLentz Yeah I couldn't think of a great schema of then deciding how to selectively load each JSON, but we can always add it in

PR #9, opened 2017-09-13
DOC: Deploy documentation via doctr

teddyrendahl:
Push to https://github.com/lcls-pcds/skywalker-docs on builds via `doctr`


hhslepicka:
Let it fly!

PR #10, opened 2017-09-13
FIX: Use NSLSII doctr now ...

teddyrendahl:
See drdoctr/doctr#190


teddyrendahl:
Merging this myself...

PR #16, opened 2017-10-17
LOG: Small fix

ZLLentz:
Switch info to debug, use module logger instead of root logger.
Largely a selfish PR, this log message spams me while loading devices in skywalker.


teddyrendahl:
Yes.

Issue #23, opened 2017-11-22
Documentation


teddyrendahl:
Update docs for 1.0.0 features. Include a much nicer README

Issue #19, opened 2017-11-22
PEP8


teddyrendahl:
`happi` should conform entirely with PEP8

PR #25, opened 2017-11-27
ENH: Device loading

### LINKED ISSUE #20: Standardize new device loading attributes
Instead of counting on our device containers to have the same name as a pcdsdevice, each device should have `device class` , `args`, and `kwargs`. This will be used to quickly load the device regardless of package

### LINKED ISSUE #21: Loading Routine
Standardized loading of multiple devices. Exceptions will be caught and reported.

### LINKED ISSUE #22: Use jinja2 templating
For devices that need to repeat information in multiple places. Allow users to supply jinja2 templates that use other device attributes

teddyrendahl:
### #20
Devices now have the following standard pieces of EntryInfo: `device_class`, `args` and `kwargs`. This divorces us from the model of a one to one mapping of `happi.Containers` to `pcdsdevices`. It also allows the addition of initialization arguments without adding EntryInfo or changing classes. By default, the `args` and `kwargs` default to the standard `ophyd.Device` (prefix, name=name) loading scheme.

### #21
Because we don't want to duplicate information in the `args` and `kwargs` that is contained in other EntryInfo. I added a function that will render `jinja2` templates with a devices specific attributes. That means for instance the defaults args are will be `"{{prefix}}"`

The awkward thing about this setup is you still want to enforce a non-string type. We need to render the template, then reconvert it using the `EntryInfo.enforce`. For instance, if your kwargs looked like:
```python
kwargs = {'z': '{{z}}'}
```
The `fill_template` command would sub-in the z value, but then would be swapped back to a float later. It is unclear how to properly implement this in the case where there are multiple attributes in a template that disagree on type. I'm open to discussing an alternative to `jinja2` templates but I want to avoid inventing our own language for doing this.

### #22
There are now two functions to support device loading

#### `from_container`
Looks at `device_class` imports the module, fills any `jinja2` templates with device information and loads the device. Exceptions **are not** caught during this function

#### `load_devices`
Load a series of devices, catching all exceptions and making sure that one bad apple will not ruin the barrel. Right now they are captured in a common `NameSpace`. The function has a `pprint` option that has a nice print out of what devices loaded and which did not. 

screen shot 2017-11-27 at 10 28 22 am

The only controversial portion is if a device raises an exception, I store that under the device name. The nice thing is you can then in the same IPython terminal examine the traceback without needing to check logs e.t.c. You also don't keep slamming tab for a device you think should be there. We can decide if this is a wanted feature.

### #19 
Cleaned up PEP8 errors using `flake8` in files I touched. This will be an ongoing effort

## Motivation
Closes #20
Closes #21 
Closes #22

## Documentation
Docstrings are written. I want to do an overhaul of the `happi` docs. So the new functons will be included as part of #23 

## Tests
New tests added for generic device loading and jinja2 templating

### DIFF
--- a/happi/device.py
+++ b/happi/device.py
@@ -94,7 +94,7 @@ class EntryInfo(object):
         -------
         value :
             Identical to the provided value except it may have been converted
-            to the correc type
+            to the correct type
 
         Raises
         ------
--- a/happi/device.py
+++ b/happi/device.py
@@ -1,23 +1,24 @@
-from __future__ import print_function
-
-###################
-# Standard Packages
-###################
+############
+# Standard #
+############
 import re
 import six
 import sys
 import logging
 
 from collections import OrderedDict
-from prettytable import PrettyTable 
+from prettytable import PrettyTable
 
-###################
-# Module Packages
-###################
+###############
+# Third Party #
+###############
+
+##########
+# Module #
+##########
 from .errors import ContainerError
 
 logger = logging.getLogger(__name__)
-
 
 
 class EntryInfo(object):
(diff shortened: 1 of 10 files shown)


happi/device.py
```diff
@@ -1,25 +1,26 @@
...
 import sys
 import logging
 
 from collections import OrderedDict
-from prettytable import PrettyTable 
+from prettytable import PrettyTable
+
+###############
+# Third Party #
+###############
```
ZLLentz: It's not a huge deal, but can we stop using these comment headers? It should be understood that it always goes standard imports, blank line, third party imports, blank line, module imports. I think this is in one of the pep style guides.

happi/device.py
```diff
@@ -63,50 +64,65 @@ class MyDevice(Device):
...
+
+        Parameters
+        ----------
+        value
+
+        Returns
+        -------
+        value :
+            Identical to the provided value except it may have been converted
+            to the correc type
```
ZLLentz: typo

happi/device.py
```diff
@@ -237,77 +247,70 @@ class Device(six.with_metaclass(InfoMeta, object)):
...
+                       enforce=bool, default=True)
+    system = EntryInfo('The system the device is involved with, i.e '
+                       'Vacuum, Timing e.t.c',
+                       enforce=str)
+    macros = EntryInfo("The EDM macro string asscociated with the "
+                       "with the device. By using a jinja2 template, "
+                       "this can reference other EntryInfo keywords.",
+                       enforce=str)
+    parent = EntryInfo('If the device is a component of another, '
+                       'enter the name', enforce=str)
```
ZLLentz: Thank you for fixing the spacing here

happi/device.py
teddyrendahl: Yeah. I can stop in future files.

PR #28, opened 2017-12-11
Cleanup

### LINKED ISSUE #19: PEP8

teddyrendahl:
## Description

A bunch of fairly small, but necessary changes to make things a little nicer.
* The entire library satisfies PEP8 requirements
* The tests no longer strictly require `mongomock`
* No longer need to import `six` as a requirement
* Documentation Cleanup

## Motivation and Context

Closes #19 

## How Has This Been Tested?

Same tests, still passing

### DIFF
--- a/docs/source/device.rst
+++ b/docs/source/device.rst
@@ -42,42 +42,17 @@ Device container. This section documents the entries when they either do not
 exist in the generic device or require further clarification on a case-by-case
 basis.
 
-Gate Valve
-++++++++++
-.. autoclass:: happi.containers.GateValve
+.. autosummary::
+   :toctree: generated
 
-Slits
-+++++
-.. autoclass:: happi.containers.Slits
-
-PIM
-+++
-.. autoclass:: happi.containers.PIM
-
-IPM
-+++++
-.. autoclass:: happi.containers.IPM
-
-Attenuator
-++++++++++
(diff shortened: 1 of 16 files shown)


docs/source/device.rst
```diff
@@ -42,42 +42,17 @@ Device container. This section documents the entries when they either do not
...
-.. autoclass:: happi.containers.PulsePicker
-
-LODCM
-+++++
-.. autoclass:: happi.containers.LODCM
-
-MovableStand
-+++++++++++++
-.. autoclass:: happi.containers.MovableStand
+.. autosummary::
```
ZLLentz: Sphinx keyword of the year right here

happi/client.py
```diff
@@ -10,7 +10,6 @@
 ###############
 # Third Party #
 ###############
```
ZLLentz: This is why I don't like the import comment headers. Look at the big "we don't use 3rd party modules" banner it leaves. https://www.python.org/dev/peps/pep-0008/#imports specifies only a blank line between each import group.

docs/source/device.rst
teddyrendahl: Yup!

happi/client.py
teddyrendahl: I'll remove these in the next PR

PR #31, opened 2017-12-22
API: Happi Restructuring

### LINKED ISSUE #17: Update documentation for new beamline naming scheme
I updated the naming conventions for the beamlines, this should be reflected in the documentation on the `Conventions` page.

### LINKED ISSUE #23: Documentation

teddyrendahl:
## Description

Large restructuring of the package that needed to get done before `1.0.0`. This PR should encompass all the major API breakages.

### Backends
Before all the backends were placed in one file; `happi/backends.py`. This meant that in order to run `happi` you needed **all** the dependencies for all the possible backends. This isn't a huge deal right now, but as `happi` adds more backends in the future this was not sustainable. There is now a happi/backends folder. The `__init__` checks for a `HAPPI_BACKEND` environment variable to see if you specify a specific type, otherwise `JSON` is used as it is more ubiquitous (and what is on the floor now). You can always still manually load a backend and give it to the client if you so choose, the default backend just gives you a littler shortcut to interpret the `Client` kwargs:

```python 
      #JSON configured as default backend
      happi.Client(path='path/to/my_json.json)

      #JSON not-configured as default backend
      happi.Client(database=JSONBackend(path='path_to_my_json.json))
```

### Client API
Formerly the `happi.Client` had a method `load_device`. This found a single matching entry and returned it. Now that we are actually loading external objects in this library, not just `happi.Devices` I thought this syntax would be very confusing. I renamed the method to `find_device`. I thought this make it clear that you will not be returned a `pcdsdevices.Device`. We can decide if we want to put a hook on the `Client` called load_device that **will** return you the instantiated object at a later date

### Tests
I updated the test fixtures to be much cleaner. In fact, the way we were testing the `mongoDB` backend was covering up a glaring bug. These are now implemented in a nicer way. I also made sure that the testing library can run without all of the optional dependencies (you shouldn't need `mongomock` to test the `simplejson` library)

### Documentation
I rearranged and updated the docs. I like the model of keeping the API and the tutorials separate. Also did a lot of work on the README. This repo might be a silent obsession: https://github.com/matiassingers/awesome-readme

Closes #17 
Closes #23 

## How Has This Been Tested?

Same tests, different fixtures

Documentation updated and rearranged
<!--


happi/client.py
```diff
@@ -174,7 +162,7 @@ def add_device(self, device):
         logger.info('Device %r has been succesfully added to the '
                     'database', device)
 
-    def load_device(self, **post):
+    def find_device(self, **post):
```
ZLLentz: If you want to make it really explicit that this looks up a database entry and doesn't actually return a `Device` subclass, you can pick a name more like `find_entry` or `find_data`, or something similar.

teddyrendahl: Yeah the awkward part is that I have a `find_document` method that returns the raw dictionary information. We can argue how useful that is ...

PR #33, opened 2018-01-22
Questionnaire Backend

### LINKED ISSUE #30: Questionnaire Backend
## Expected Behavior

Have a happi Client that connect directly to the LCLS Questionnaire and provide a read-only inteface

## Current Behavior

No support for Questionnaire loading. This would currently have to be dumped to JSON.

## Possible Solution

Hopefully this can be an in-memory JSONBackend with no save method.

teddyrendahl:
## Description

Backend database type for the Questionnaire. 

### Usage
``` python
import happi
from happi.backends.qs_db import QSBackend
client = Client(database=QSBackend(15, 'LP69'))
client.all_devices
```
Prints:
```
[Device sam_x (prefix=XPP:USR:MMS:01, z=-1.0),
 Device sam_z (prefix=XPP:USR:MMS:02, z=-1.0),
 Device sam_y (prefix=XPP:USR:MMS:03, z=-1.0),
 Device sam_r (prefix=XPP:USR:MMS:04, z=-1.0),
 Device sam_az (prefix=XPP:USR:MMS:05, z=-1.0),
 Device sam_flip (prefix=XPP:USR:MMS:06, z=-1.0)]
```

### Details
* Adds a few dependencies involved with the QuestionnaireClient. These are fully optional.
* The `QSBackend` inherits from the `JSONBackend`. The reason is the `JSONBackend` already has all the functionality to keep the devices organized in a dictionary. The `QSBackend` uses the same functions but just never dumps.
* The `QSBackend` is completely read-only
* Currently the `QSBackend` needs to be re-initialized to load new information from the questionnaire. This is to avoid repeated requests to get redundant information. 
* Dropped `3.4` support. This is because we are only building the questionnaire requirements in our conda channel for `3.5` and `3.6`
 
## Motivation and Context

Closes #30 

## How Has This Been Tested?

Created a mock Questionnaire client that returns a sample dictionary dump. Did not use the actual questionnaire to avoid kerberos authentication and other strangeness

Added to backend documentation


happi/backends/qs_db.py
```diff
@@ -0,0 +1,151 @@
...
+
+from psdm_qs_cli import QuestionnaireClient
+
+from .json_db import JSONBackend
+from ..errors import DatabaseError
+
+logger = logging.getLogger(__name__)
+
+
+class QSBackend(JSONBackend):
```
ZLLentz: I'll approve without any changes here, but it might be better organizationally to have `QSBackend` and `JSONBackend` both inherit from a shared parent rather than from each other. The code will be easier to maintain if the `QSBackend` class defines only how to get data from the Questionnaire and the `JSONBackend` defines only how to get data from a json file, with some other class defining the shared features that these and similar backends would need.

happi/tests/conftest.py
```diff
@@ -23,6 +23,19 @@
 requires_mongomock = pytest.mark.skipif(not has_mongomock,
                                         reason='Missing mongomock')
 
+# Conditional import of mongomock
```
ZLLentz: This comment doesn't make sense, it was probably copy/pasted

teddyrendahl:
Yeah. It would be nice if the dictionary was nested a little nicer. I think if we decide that we like this paradigm we should either work with Murali or build more smarts into the `QuestionnaireClient` itself. 

For now, I think this will be stable enough to see if SRD likes the idea. Then as we add more complexity, we can decide where to put the extra logic (avoiding the happi client itself I think)

happi/backends/qs_db.py
teddyrendahl: I agree. I think that's the correct way. I'll mark it as an issue for later review.

Issue #36, opened 2018-01-26
Container 


teddyrendahl:
## Expected Behavior

We should not maintain a list of container with hard-coded paths to devices. Instead we should supply useful helpful functions to turn objects/classes into happi devices. Much of the manual labor can be avoided avoided by inspecting the class itself.

Example Syntax
```python
container = container_from_class(pcdsdevices.device_types.GateValve)
happi_client.add_device(container(prefix='Tst:VGC:01', name='Easy Valve'))
container.name == 'Easy Valve'
container.device_class == 'pcdsdevices.device_types.GateValve'
```

klauer:
Related to #56. Implementation may depend on registry decision made in #126.

PR #37, opened 2018-01-26
MAINT: Update pcdsdevices Containers

### LINKED ISSUE #27: Update containers to point to pcdsdevices
## Expected Behavior

We are no longer expecting to explicitly match the name of a container to a pcdsdevice. We should have the defaults of these classes point to the correct class using the `device_class`. Each should also have the correct `args` and `kwargs` formulated to pipe EntryInfo names into the device constructor.

## Current Behavior

Specialized containers have `None` set for `device_class`

teddyrendahl:
## Description

Update containers to load proper `pcdsdevices` classes. This way we know which are the required EntryInfo parameters, and you do not have to repeatedly point all your added devices to the same locations in `pcdsdevices`.

This is a temporary fix until #36 has a solution.

## Motivation and Context

Closes #27 

## Tests
Did not add tests to avoid a`pcdsdevices` dependency.

### DIFF
--- a/happi/containers.py
+++ b/happi/containers.py
@@ -6,19 +6,6 @@ from copy import copy
 from .device import Device, EntryInfo
 
 
-# Broad device categories
-class MPS(Device):
-    """
-    Parent class for devices that are in MPS.
-    """
-    system = copy(Device.system)
-    system.default = 'mps'
-    mps = EntryInfo('MPS PV associated with the Device',
-                    optional=False, enforce=str)
-    veto = EntryInfo('Whether MPS considers this a veto device',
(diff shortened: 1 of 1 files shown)


happi/containers.py
```diff
@@ -98,6 +85,8 @@ class GateValve(Vacuum, MPS):
     """
     prefix = copy(Vacuum.prefix)
     prefix.enforce = re.compile(r'.*VGC.*')
+    device_class = copy(Device.device_class)
+    device_class = 'pcdsdevices.device_types.GateValve'
```
ZLLentz: Should this be `device_class.default = `

PR #39, opened 2018-01-30
Exception Details

teddyrendahl:
## Description

1. Was having trouble deciphering an exception in the happi Client. Added the `raise . from .` syntax to give a more descriptive traceback. I figured this should just go in the library permanently.
2. While writing tests for another module, I discovered that you can not use `load_devices` if one of your classes references metadata on load. This is because we try and enforce the type by looking up the EntryInfo, but there is no EntryInfo for metadata. This just wraps it in a try / except block and trusts that the given metadata is already the correct type.

### DIFF
--- a/happi/client.py
+++ b/happi/client.py
@@ -189,11 +189,11 @@ class Client:
         logger.debug("Instantiating device based on found information ...")
         try:
             device = self.create_device(doc['type'], **doc)
-        except (KeyError, TypeError):
+        except (KeyError, TypeError) as exc:
             raise EntryError('The information relating to the device class '
                              'has been modified to the point where the object '
                              'can not be initialized, please load the '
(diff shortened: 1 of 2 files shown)


happi/client.py
```diff
@@ -189,11 +189,11 @@ def find_device(self, **post):
         logger.debug("Instantiating device based on found information ...")
         try:
             device = self.create_device(doc['type'], **doc)
-        except (KeyError, TypeError):
+        except (KeyError, TypeError) as exc:
             raise EntryError('The information relating to the device class '
                              'has been modified to the point where the object '
                              'can not be initialized, please load the '
-                             'corresponding document')
+                             'corresponding document') from exc
```
ZLLentz: Forgot this was a thing, I like it a lot

PR #45, opened 2018-02-09
ENH: Cache

### LINKED ISSUE #29: Device Caching
## Expected Behavior

The device loading should keep track of which devices it has loaded so far. This way we do not attempt to load the same device twice. 

## Current Behavior

No caching done.

## Possible Solution

Have a `happi.cache` that has a mapping of loaded devices to containers. This way it is possible to go find the container information for a prior device. Also if a device is requested twice we just return a reference to the first one.

teddyrendahl:
## Description

Simplest implementation of device caching. Idea is to check if a device has already been created rather than creating a new one. Cache is available as `happi.cache`. If a user wants to force a new object to be created they can pass `use_cache=False` into `from_container`. This creates a new device and replaces it in the cache.

### Implementation Detail
* This is currently just a regular dictionary. We can make it a `weakref` if we want to destroy devices when they go out of scope.

## Motivation and Context

Closes #29 

## How Has This Been Tested?

Added a test where we compare  object memory addresses using  `id` 

 Added a docstring in `from_container`.

### DIFF
--- a/happi/loader.py
+++ b/happi/loader.py
@@ -134,7 +134,7 @@ def from_container(device, attach_md=True, use_cache=True):
     return obj
 
 
-def load_devices(*devices, pprint=False, namespace=None):
+def load_devices(*devices, pprint=False, namespace=None, **kwargs):
     """
     Load a series of devices into a namespace
 
@@ -149,6 +149,9 @@ def load_devices(*devices, pprint=False, namespace=None):
     namespace : obj, optional
         Namespace to collect loaded devices in. By default this will be a
         ``types.SimpleNamespace``
+
+    kwargs:
+        Are passed to :func:`.from_container`
     """
     # Create our namespace if we were not given one
     namespace = namespace or types.SimpleNamespace()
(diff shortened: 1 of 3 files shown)


happi/loader.py
```diff
@@ -72,10 +74,22 @@ class is then instantiated with the given args and kwargs provided.
...
+        needed, set `use_cache` to False and a new object will be created,
+        overriding the current cached object
+
     Returns
     -------
     obj : happi.Device.device_class
     """
+    # Return a cached version of the device if present and not forced
+    if use_cache and device.prefix in cache:
+        return cache[device.prefix]
```
ZLLentz: Devil's advocate for a second: will two devices ever have the same prefixes? I can't immediately think of any cases, but it's good to worry. If this is a fundamental assumption of `happi` that's totally good (is it the `_id` field?), I just want to make sure.

happi/tests/test_loader.py
```diff
@@ -28,9 +28,20 @@ def test_from_container():
...
 
+def test_caching():
+    # Create a datetime device
+    d = TimeDevice(name='Test', prefix='Tst:This:2', beamline='TST',
+                   device_class='datetime.timedelta', args=list(), days=10,
+                   kwargs={'days': '{{days}}', 'seconds': 30})
+    td = from_container(d)
+    assert d.prefix in cache
+    assert id(td) == id(from_container(d))
+    assert id(td) != id(from_container(d, use_cache=False))
```
ZLLentz: 👍

happi/loader.py
teddyrendahl: Yeah. `happi` assumes that you won't ever have two devices with the same prefix. I thought about making it a hash of name and prefix, but decided we shouldn't allow two devices with the same prefix.

PR #51, opened 2018-03-02
ENH:Questionnaire Motor Guess

teddyrendahl:
DO NOT MERGE UNTIL https://github.com/pcdshub/pcdsdevices/pull/167

## Description

When loading motors from the Questionnaire, guess which type each is by looking at the prefix. I view this is a temporary fix until the QuestionnaireClient can tell us explicitly which class should be loaded.

**Question:** Should the default be `pcdsdevices.epics_motor.PCDSMotorBase` or `ophyd.EpicsMotor`? Not sure ... My only thought is that our base version will work on true MotorRecord motors, but the `EpicsMotor` won't work with our bastards.

## Motivation and Context

Without this fix Newport and PMC100 motors will not load properly from the Questionnaire. 

## How Has This Been Tested?

Added tests for both guessing the motor type, and that it works in situ.

### DIFF
--- a/happi/backends/qs_db.py
+++ b/happi/backends/qs_db.py
@@ -29,7 +29,8 @@ def guess_motor_class(prefix):
     Returns
     -------
     device_class : str
-        Type of EpicsMotor. If not, we assume can use pcdsd
+        Type of EpicsMotor. If not, we assume
+        `pcdsdevices.epics_motor.PCDSMotorBase`
     """
     for _typ in motor_types:
         if _typ in prefix:
--- a/happi/backends/qs_db.py
+++ b/happi/backends/qs_db.py
@@ -12,6 +12,31 @@ from ..errors import DatabaseError
 logger = logging.getLogger(__name__)
 
 
+# Declare our motor types
+motor_types = {'MMS': 'pcdsdevices.epics_motor.IMS',
+               'MMN': 'pcdsdevices.epics_motor.Newport',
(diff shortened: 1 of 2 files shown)


happi/backends/qs_db.py
```diff
@@ -12,6 +12,31 @@
...
+    ----------
+    prefix : str
+
+    Returns
+    -------
+    device_class : str
+        Type of EpicsMotor. If not, we assume can use pcdsd
+    """
+    for _typ in motor_types:
+        if _typ in prefix:
```
ZLLentz: Might want to be more robust than this, just in case a motor comes through with a wonky name like `XPP:USR:MZM:AMMN`... contrived example, I know. I suggest checking `prefix.split(':')` and trying to match a piece of it.

ZLLentz: Actually, forget about this, we have to strip this out when the questionnaire gets updated anyway.

ZLLentz:
Just make sure you take care of the flake8 issue that travis is mad about 👍

teddyrendahl:
Okay one more thought. This will break for anything older than the current `pcdsdevices`. These classes were just introduced but never actually imported. How do I express this implicit dependency?

ZLLentz:
I think if we've committed to making this dependency implicit, there's nothing to explicitly do about it. Libraries that need an older `pcdsdevices` must also specify an older `happi`.

PR #67, opened 2018-07-23
ENH: Threads

### LINKED ISSUE #66: Port Threaded Loading from hutch_python
There are quite a few applications that will want to load a large number of devices from happi. It would be nice to batch and thread loading where we can.

ZLLentz:
Closes #66 

## Description

- `load_devices` has an option to create each device in a background thread
- `load_devices` has an option to call an arbitrary function on the loaded devices in the background thread, to be used to wait for connection, etc.

## Motivation and Context

- This is needed before we transfer the camera load from `hutch-python` to `happi`
- This allows us to efficiently load devices that need PV gets to initialize that could time out
- This allows us to catch timeouts/disconnections early without sacrificing tons of time for applications like `lightpath`, but also give us flexibility to add checks in applications that need specific checks without thinking of all the possibilities in the `happi` module. It also helps stave off the "hidden" `ophyd` dependence by not assuming we only have `Device` and `Signal` objects.

## How Has This Been Tested?

- Added testing
- Works for xpp

- Expanded docstrings appropriately

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
+.pytest_cache
 nosetests.xml
 coverage.xml
 *,cover
--- a/happi/loader.py
+++ b/happi/loader.py
@@ -172,8 +172,8 @@ def from_container(device, attach_md=True, use_cache=True, threaded=False):
     return obj
 
 
-def load_devices(*devices, pprint=False, namespace=None, threaded=True,
-                 **kwargs):
+def load_devices(*devices, pprint=False, namespace=None, use_cache=True,
+                 threaded=True, **kwargs):
     """
     Load a series of devices into a namespace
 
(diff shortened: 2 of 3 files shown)


ZLLentz:
A little more to do here: disconnected devices should still be returned, with a log message indicating such. They might reconnect later.

klauer:
I think this looks good overall. 

For safety's sake we might need to add some more complexity to determine which devices get instantiated in which threads. Specifically, we don't want ophyd (-> pyepics) to race in creating a PV instance. Multiple devices may share the same PV, after all.

ZLLentz:
`Multiple devices may share the same PV, after all.`

This makes me upset for some reason

klauer:
@ZLLentz I agree for the most part. Maybe one day we could restrict that ability in our pcdsdevices `Device` subclass at some point 🤷‍♂

PR #70, opened 2018-10-09
Minor Fixes

teddyrendahl:
## Description

* `pip` requirements were a little out of date
* `JSONBackend` watches to see if you are going to overwrite an existing database. The fear is that someone types `initialize=True` and erases their whole db. However, we shouldn't be scared of writing to an empty file.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,6 +3,7 @@ codecov
 doctr
 ipython
 matplotlib
+pymongo
 sphinx
 sphinx_rtd_theme
 git+https://github.com/slaclab/psdm_qs_cli.git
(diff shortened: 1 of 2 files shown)


dev-requirements.txt
```diff
@@ -3,7 +3,8 @@ codecov
 doctr
 ipython
 matplotlib
+pymongo
 sphinx
 sphinx_rtd_theme
-psdm_qs_cli
+git+https://github.com/slaclab/psdm_qs_cli.git
```
ZLLentz: What does this do? If it checks it out from git when you load dev-requirements then this isn't good practice.

ZLLentz: Actually this completely stalled the travis on closer inspection

PR #71, opened 2018-10-09
Discussion: Explicit import of from_container

### LINKED ISSUE #72: Documentation Keyword
## Expected Behavior

Add a new default keyword to `happi` to store relevant documentation

teddyrendahl:
## Description

We didn't have the top level import for loading a single device -> `from_container`.

I kind of realized that this should be a default method on `Client`. I think the main reason I didn't is because I couldn't think of  a name. Thoughts on?

`find_device` -> search and return `happi.Device`
`load_device` -> search and return `ophyd.Device`

I think the reason I didn't do this before is in `happi` < `1.0.0`, `load_device` **was** a function that returned a `happi.Device`. I renamed it to avoid future confusion. 

Closes #72

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,6 @@ script:
   - |
     if [[ -n "$DOCTR_DEPLOY_ENCRYPTION_KEY_PCDSHUB_HAPPI" && $BUILD_DOCS ]]; then
       pushd docs
-      sphinx-autogen -o source/generated source/*.rst 
       make html
       popd
       # Publish docs.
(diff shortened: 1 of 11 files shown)


ZLLentz:
Your function names are fine. A more fundamental problem I realized when trying to decipher this with morning brain is that `happi.Device` should have been called something more descriptive due to the conflicts with `ophyd.Device`. Maybe too late to fix that.

teddyrendahl:
Went through with this. Going to use this as an excuse to dive into improving documentation...

rajanplumley:
> Added docs. Would appreciate if someone read through the docs real quick.

Docs look good, excited to try `load_device` out.

PR #75, opened 2018-10-18
Additional Keywords

### LINKED ISSUE #58: Lightpath Keyword
## Expected Behavior

In the near future we will like to add devices that our "active" but do not get included in the lightpath. We can either add a "lightpath=True" default to Device or we `lightpath` just needs to be aware of the device classes it needs.

### LINKED ISSUE #72: Documentation Keyword

teddyrendahl:
## Description

Adds two new keywords:

**lightpath** - Easy key that makes it very transparent whether we should include this device in the LCLS Lightpath. I like this way better than keeping a big list of accepted classes. `active` always overrules this (i.e an inactive device with `ligthpath=True` will not be shown

**documentation** - Keeping this very ambiguous. Not sure what I really want it to look like in the end. It could be a string that points to a single url or a list of sources or a dictionary that has a keyword for each documenting source. This is also slightly awkward as what you probably want is to have this on a class basis, not a device? Maybe not? Not sure.. In any case it wouldn't be hard to write a little script:

```python
my_awesome_html ='www.wtf_is_wrong_with_this_motor.com'
for device in client.search(device_class='pcdsdevices.device_types.IMS'):
     if my_awesome_html not in device.documentation:
          device.documentation.append(my_awesome_html)
          device.save()
```

## Motivation and Context

Closes #58 
Closes #72

### DIFF
--- a/happi/device.py
+++ b/happi/device.py
@@ -272,6 +272,7 @@ class Device(metaclass=InfoMeta):
                        'enter the name', enforce=str)
     lightpath = EntryInfo("If the device should be included in the ",
                           "LCLS Lightpath", enforce=bool, default=False)
+    documentation = EntryInfo("Relevant documentation for the Device")
 
     def __init__(self, **kwargs):
         # Load given information into device class
--- a/happi/device.py
+++ b/happi/device.py
@@ -270,7 +270,7 @@ class Device(metaclass=InfoMeta):
                        enforce=str)
     parent = EntryInfo('If the device is a component of another, '
                        'enter the name', enforce=str)
-    lightpath = EntryInfo("If the device should be included in the ",
+    lightpath = EntryInfo("If the device should be included in the "
                           "LCLS Lightpath", enforce=bool, default=False)
     documentation = EntryInfo("Relevant documentation for the Device")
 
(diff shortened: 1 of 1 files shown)


happi/device.py
```diff
@@ -270,6 +270,9 @@ class Device(metaclass=InfoMeta):
                        enforce=str)
     parent = EntryInfo('If the device is a component of another, '
                        'enter the name', enforce=str)
+    lightpath = EntryInfo("If the device should be included in the ",
```
klauer: Stray comma at the end?

happi/device.py
```diff
@@ -270,6 +270,9 @@ class Device(metaclass=InfoMeta):
...
+                          "LCLS Lightpath", enforce=bool, default=False)
+    documentation = EntryInfo("Relevant documentation for the Device")
```
klauer: What's the implication of not using `enforce` here?

teddyrendahl: Yeah. It doesn't really leave the opportunity for multiple documentation sources, but you could argue that it should just point at a documentation homepage and child pages are linked from there.

I'll put it as a string if no one else here has opinions

PR #76, opened 2018-10-19
Motor

### LINKED ISSUE #57: Motor Container
## Expected Behavior

We need a container for basic motors. This will probably start referring to `PCDSMotorBase` and require a more specific setting from the operator.

teddyrendahl:
## Description

Add a generic `Motor` container. By default this points to the factory created in https://github.com/pcdshub/pcdsdevices/pull/306. 

There was also some code that in the Questionnaire Backend that essentially had the same logic as the factory. This is obviously the wrong place to put this as if we scramble around our motor classes we need to remember the 

## Motivation and Context

Closes #57 

Added to autosummary

### DIFF
--- a/happi/tests/test_backends.py
+++ b/happi/tests/test_backends.py
@@ -157,9 +157,3 @@ def test_qsbackend_with_client(mockqsbackend):
     assert len(c.all_devices) == 6
     assert all([d.device_class == 'pcdsdevices.epics_motor.IMS'
                 for d in c.all_devices])
-
-
-@requires_questionnaire
-def test_guess_motor_class():
-    from happi.backends.qs_db import guess_motor_class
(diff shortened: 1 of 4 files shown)


happi/tests/test_backends.py
```diff
@@ -155,11 +155,5 @@ def test_qs_find(mockqsbackend):
 def test_qsbackend_with_client(mockqsbackend):
     c = Client(database=mockqsbackend)
     assert len(c.all_devices) == 6
-    assert all([d.device_class == 'pcdsdevices.epics_motor.IMS'
+    assert all([d.device_class == 'pcdsdevices.device_types.Motor'
```
ZLLentz: It's confusing what you're testing in here because `Motor` isn't really a class

PR #77, opened 2018-10-19
Python 3.7

teddyrendahl:
## Description

Starting the train! One small `regex` issue which is slightly annoying but it was a private variable...


teddyrendahl:
Only concern here is having to change the distro just to build `3.7` on Travis. If you guys don't care I don't... Just wanted to make you guys aware.

PR #80, opened 2018-11-15
Screen setup

### LINKED ISSUE #79: Update screen storage information
## Expected Behavior

Change the fields that specify screens to be:
* `display_screen`
* `embedded_screen`
* `engineering_screen`

I also want to delete `macros`. For L2S-I just the `happi` object itself can be passed in as JSON as the `macros` 😄 

## Current Behavior

Tailored for EDM

teddyrendahl:
## Description

For `L2S-I` each device will have three screens. I have adjusted the `happi.Device` to have three fields for these. 

Closes #79

### DIFF
--- a/happi/device.py
+++ b/happi/device.py
@@ -1,6 +1,7 @@
 import re
 import sys
 import logging
+import warnings
 
 from collections import OrderedDict
 from prettytable import PrettyTable
@@ -359,6 +360,13 @@ class Device(metaclass=InfoMeta):
             post.update(self.extraneous)
 
         return post
+
+    @property
+    def screen(self):
+        warnings.warn("The 'screen' keyword is no longer used in Happi as it "
+                      "lacks specificity. Use one of detailed_screen, "
+                      "embedded_screen, or engineering screen instead")
+        return self.detailed_screen
 
     def save(self):
         """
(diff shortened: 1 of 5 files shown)


happi/device.py
```diff
@@ -264,8 +265,14 @@ class Device(metaclass=InfoMeta):
                        enforce=dict, default={'name': '{{name}}'})
     stand = EntryInfo('Acronym for stand, must be three alphanumeric '
                       'characters', enforce=re.compile(r'[A-Z0-9]{3}$'))
-    screen = EntryInfo('The absolute path to the main control screen',
-                       enforce=str)
+    detailed_screen = EntryInfo('The absolute path to the main control screen',
+                                enforce=str)
+    embedded_screen = EntryInfo('The absolute path to the '
```
klauer: Is absolute path the best way to do things here?

Alternatively, should there be a `screen_root` and these become relative to that?

teddyrendahl: I think when we decide on a convention we could have this (not just for a device, but globally). Until then I'd like to keep each entry as specific as possible.

teddyrendahl:
Nothing is ever easy. My 15 line PR doubled in size:

1. I broke a test (that was on me)
2. `flake8` updated at some point so had a few style complaints floating around
3.  After testing locally again, I realized I wasn't being very kind to the environment variables required for the tests. I know make sure to cache the original values and restore them at the end.

happi/device.py
klauer: OK, seems reasonable.

PR #81, opened 2018-11-18
DEV: use requirements.txt to fill in install_requires

tacaswell:


teddyrendahl:
This should work ... Travis is not happy though. I'll investigate.

PR #82, opened 2018-11-20
Windows Import

teddyrendahl:
## Description

* `fcntl` is not importable on Windows. We do not plan to support or test on Windows, but the package should at least be importable and there isn't a real fundamental reason that the `JSON` backend shouldn't work on any operating system. 
* The testing suite was kind of a mess. We were using fixtures improperly, calling them as regular functions. In `pytest>=4.0` this now explodes. This is now handled better, using a more elegant fixture pattern.

Attn. @tacaswell

### DIFF
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -15,11 +15,9 @@ logger = logging.getLogger(__name__)
 
 try:
     import fcntl
-    on_unix = True
 except ImportError:
     logger.warning("Unable to import 'fcntl'. Will be unable to lock files")
-    on_unix = False
-
+    fcntl = None
 
 class JSONBackend(metaclass=Backend):
     """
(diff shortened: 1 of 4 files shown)


happi/backends/json_db.py
```diff
@@ -14,6 +13,14 @@
 logger = logging.getLogger(__name__)
 
 
+try:
+    import fcntl
+    on_unix = True
```
klauer: 👍 but you might consider saying `if fcntl is not None` and not having an extra state variable

happi/tests/conftest.py
```diff
@@ -51,8 +51,8 @@ def device_info():
 
 
 @pytest.fixture(scope='function')
-def device():
-    t = Device(**device_info())
+def device(device_info):
```
klauer: Nice pattern! 😄

happi/backends/json_db.py
teddyrendahl: 🤔 makes sense

PR #84, opened 2019-01-14
Cmd line tool

jsheppard95:
## Description
Added command line tool as python script (`happi/happi_cl.py`) which can add and search for existing devices in the happi database. Also fixed bug where `client.find_device()` returned the first database entry when no arguments were given.

Currently the command line script allows the user to search and add devices as shown below, but the current way I add devices is messy/highly dependent on the format of the user input (especially specifying `args` and `kwargs`) and I could use some feedback to come up with either more helper functions or a different approach.

## Motivation and Context
This change will allow for easier usage of the happi client without needing an IPython session (however a conda environment is still required before running). Instead the usage thus far would be:
`$ python happi_cl.py --help`
`usage: happi_cl.py [-h] [--search] [--field FIELD] [--value VALUE] [--add]`

`happi command line tool`

`optional arguments:`
`  -h, --help     show this help message and exit`
`  --search       search for device`
`  --field FIELD  field in happi database to consider`
`  --value VALUE  value of 'field' to return results`
`  --add          add a device`

`$ python happi_cl.py --search --field z --value 978.9`
screenshot from 2019-01-14 11-52-49

`$ python happi_cl.py --add`
screenshot from 2019-01-14 11-59-21

## How Has This Been Tested?
I am currently unsure how to use pytest to automate the testing of the command line script. The Python functions that this script calls are already tested in tests/test_client.py and tests/test_backends.py but I do need to somehow create pytest fixtures that capture the user input process from the command line.


teddyrendahl:
Okay here are my thoughts:

1. Let's wait on `add` that is complicated. 
2. For `search` you should use [nargs](https://docs.python.org/3/library/argparse.html#nargs). I would consider `+` or `args.REMINDER`. You'll still have to pair them up
3. Checkout https://github.com/pcdshub/typhon/blob/master/typhon/cli.py. I would like for you to move the command line file inside the package.
4. Checkout this in the `setup.py` https://github.com/pcdshub/typhon/blob/7d7d34aaa4fe9283483b5298bf382e15ddb3cbf5/setup.py#L21

Here is some reference -> https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

happi/cli.py
```diff
@@ -0,0 +1,110 @@
+"""
+Module for a command line interface to happi
+"""
+# cli.py
+# NOTE: Does not work from any environment, must have pcds_conda activated
+## Can this script activate/deactivate pcds_conda upon use?
```
teddyrendahl: If you use the `setup.py` that I linked, the command will be made available once you enter `pcds_conda`

happi/cli.py
```diff
@@ -0,0 +1,110 @@
...
+                    help='search for device')
+parser.add_argument('--field', help='field in happi database to consider')
+parser.add_argument('--value', help="value of 'field' to return results",
+                    type=str)
+
+parser.add_argument('--add', action='store_true',
+                    help='add a device')
+
+# TODO: Implement edit option
+#parser.add_argument('--edit', action='store_true',
```
teddyrendahl: Let's wait on this

happi/cli.py
```diff
@@ -0,0 +1,110 @@
...
+#                    help='edit a device')
+
+#args = parser.parse_args()
+#arg_dict = vars(args)
```
teddyrendahl: Make sure these get deleted

happi/cli.py
```diff
@@ -0,0 +1,110 @@
...
+
+def happi_cli(args):
+    args = parser.parse_args(args)
+    print(args)
+    arg_dict = vars(args)
+    print(arg_dict)
+    # Instantiate Client and JSONBackend
+    json_backend = json_db.JSONBackend('/home/sheppard/PCDSGitHub/device_config/db.json')
```
teddyrendahl: We don't want this path hard-coded. I would use `happi.Client.from_config`. Checkout the tests to see how to set it up yourself 😉

happi/cli.py
```diff
@@ -0,0 +1,110 @@
...
+# TODO: Implement edit option
+#parser.add_argument('--edit', action='store_true',
+#                    help='edit a device')
+
+#args = parser.parse_args()
+#arg_dict = vars(args)
+
+def happi_cli(args):
+    args = parser.parse_args(args)
+    print(args)
```
teddyrendahl: Use the `logger` here. That way you don't have to remember to delete them. Just give it a little nicer message

happi/cli.py
jsheppard95: I think I set this up, just made a commit (was working locally) and I need to try creating a new conda environment on the ps machines to see if this works

happi/cli.py
jsheppard95: I'm not sure if I did this quite right. After looking at `client.from_config` and the implementation used in the tests I thought I should create a file `happi/happi.cfg` (I believe this gets created and removed during your tests), but this contains a line pointing to my hard-coded `db.json` file. For actual deployment I figured it could point to the real location of `device_config/db.json` but that might not have been what you had in mind.

happi/cli.py
jsheppard95: I see now `arg_dict` is not necessary, removed the print statements and used as logger DEBUG messages

happi/cli.py
jsheppard95: Also I added the options for `--version` and `--verbose`. I made this commit so I could move to the ps machines and try to create a development conda package and test the `--version` option. The way I'm using `--verbose` may be slightly off, specifically at line 13:
`logger = logging.getLogger(__name__)`
I know this depends on how I run `cli.main()`

setup.py
```diff
@@ -4,14 +4,17 @@
 with open('requirements.txt') as f:
     requirements = f.read().split()
 
+git_requirements = [r for r in requirements if r.startswith('git+')]
+requirements = [r for r in requirements if not r.startswith('git+')]
+print("User must install the following packages manually:\n" +
+        "\n".join(f' {r}' for r in git_requirements))
```
ZLLentz: @teddyrendahl thoughts on skipping this print if git_requirements is empty? It makes tons of spammy prints in all the conda builds (like, this file must get run 6+ times)

teddyrendahl: Fine by me.

happi/cli.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Version endpoint
+    if args.version:
+        print(f'Happi: Version {happi.__version__} from {happi.__file__}')
+        return
+    logger.debug('Command line arguments: %r' % args)
+
+    client = happi.client.Client.from_config(cfg=args.path)
+
+    if args.search:
```
teddyrendahl: This is a good start. I think we want to support multiple search terms i.e

`happi --search beamline MFX stand DG1`

happi/cli.py
```diff
@@ -0,0 +1,67 @@
...
+        if args.search[0] is 'z':
```
teddyrendahl: Instead of checking just if it is a `z` keyword try `args.search[1].is_number()` this will tell you if the key you have could be converted to a number or not. If it looks like a number it probably is 😉

conda-recipe/meta.yaml
```diff
@@ -19,9 +19,12 @@ requirements:
       - simplejson
       - pymongo 
       - python
+      - coloredlogs
 test:
     imports:
       - happi
+    commands:
```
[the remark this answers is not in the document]
jsheppard95: Yes I'm working on another test that compares the command line output to that of just using the python functions. Also @ZLLentz showed me the `commands:` trick to test the installation, just to give credit where its due.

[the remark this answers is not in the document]
jsheppard95: Also just made a new test per our discussion @teddyrendahl and I'll now go through these comments/make a new commit

happi/cli.py
```diff
@@ -54,12 +54,19 @@ def happi_cli(args):
...
-            args.search[1] = float(args.search[1])
-        search_args = {args.search[0]: args.search[1]}
+        for i in range(len(args.search)):
+            if args.search[i] is 'z':
+                args.search[i + 1] = float(args.search[i + 1])
 
-        device = client.find_device(**search_args)
-        device.show_info()
+        search_args = {}
+        for i in range(0, len(args.search), 2):
```
teddyrendahl: Maybe check if `len(args.search)` is an even number? Otherwise this will raise an `IndexError`

happi/cli.py
```diff
@@ -0,0 +1,67 @@
...
+import coloredlogs
+
+logger = logging.getLogger(__name__)
+
+# Argument Parser Setup
+parser = argparse.ArgumentParser(description='happi command line tool')
+
+parser.add_argument('--path', type=str,
+                    help='path to happi configuration file')
+parser.add_argument('--search', nargs=argparse.REMAINDER,
```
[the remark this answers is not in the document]
jsheppard95: This seems a bit difficult given the way I read in search arguments from the command line using `nargs=argparse.REMAINDER` which stores each entry following `--search` into a list where entries are separated by spaces. I could do something like `happi search beamline=MFX` and then do some string processing. I may be wrong, but I actually think `netconfig` works as follows:
`netconfig search <pattern>`
i.e
and it returns results with fields whose values (as strings) match any part of `<pattern>`.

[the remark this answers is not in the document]
klauer: Now that you've got the argparse basics down, it's probably a good opportunity to dive a bit deeper: https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers

A subparser will allow those subcommands to be implemented correctly - and nicely for the user, too.

happi/cli.py
```diff
@@ -0,0 +1,90 @@
...
+        for i in range(0, len(args.search_criteria), 2):
+            client_args[args.search_criteria[i]] = args.search_criteria[i+1]
+
+        devices = client.search(**client_args)
+        if devices:
+            for dev in devices:
+                dev.show_info()
+            return devices
+        else:
+            print('No devices found')
```
teddyrendahl: Maybe swap this to `logger.error` instead of `print`

happi/cli.py
```diff
@@ -0,0 +1,90 @@
...
+        # Ensure we have an even number of search elements
+        # Should always have key:value pairs
+        if len(args.search_criteria) % 2 is not 0:
+            raise EntryError('Search criteria should be given as key '
+                             'value pair\ni.e:\n'
+                             'happi search beamline MFX stand DG1')
+
+        # Convert any numbers from strings to floats
+        for i in range(len(args.search_criteria)):
+            if args.search_criteria[i].replace('.', '').isnumeric():
```
teddyrendahl: I'd put a `logger.debug` message here noting which values we are treating as numbers.

happi/cli.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Ensure we have an even number of search elements
+        # Should always have key:value pairs
+        if len(args.search_criteria) % 2 is not 0:
+            raise EntryError('Search criteria should be given as key '
+                             'value pair\ni.e:\n'
+                             'happi search beamline MFX stand DG1')
+
+        # Convert any numbers from strings to floats
+        for i in range(len(args.search_criteria)):
```
teddyrendahl: This can just be `for criteria in args.search_criteria`

happi/cli.py
```diff
@@ -0,0 +1,90 @@
...
+            if args.search_criteria[i].replace('.', '').isnumeric():
+                args.search_criteria[i] = float(args.search_criteria[i])
+
+        # Get search criteria into dictionary for use by client
+        client_args = {}
+        for i in range(0, len(args.search_criteria), 2):
```
teddyrendahl: Let's combine these two `for` loops

happi/client.py
```diff
@@ -186,6 +186,8 @@ def find_device(self, **post):
         device : :class:`.Device`
             A device that matches the characteristics given
         """
+        if len(post) == 0:
+            raise EntryError('No information pertinent to device given')
```
teddyrendahl: `SearchError`?

happi/cli.py
```diff
@@ -0,0 +1,90 @@
...
+    logger.debug("Happi client: %r" % client)
+    logger.debug('Happi command: %r' % args.cmd)
+
+    if args.cmd == 'search':
+        logger.debug("We're in the search block")
+
+        # Ensure we have an even number of search elements
+        # Should always have key:value pairs
+        if len(args.search_criteria) % 2 is not 0:
+            raise EntryError('Search criteria should be given as key '
```
teddyrendahl: `SearchError`?

PR #89, opened 2019-03-07
DEV: Ensure tests pass without mongomock

teddyrendahl:
## Description

More experienced @teddyrendahl knows that this too complicated a testing setup. A `pytest` fixture that has a variable number of parameter based on what you can import. 

That being said this is the one-line fix to make sure our tests still run w/o `mongomock` installed. The main reason I don't want to make this a testing requirement is it has no CONDA package and don't think the effort of packaging it ourselves is worth the time.

## How Has This Been Tested?

Verified locally without `mongomock`

### DIFF
--- a/happi/tests/conftest.py
+++ b/happi/tests/conftest.py
@@ -86,17 +86,19 @@ def mockjsonclient(device_info):
 @pytest.fixture(scope='function')
 def mockmongoclient(device_info):
     with patch('happi.backends.mongo_db.MongoClient') as mock_mongo:
-        mc = MongoClient()
-        mc['test_db'].create_collection('test_collect')
-        mock_mongo.return_value = mc
-        # Client
-        backend = MongoBackend(db='test_db',
-                               collection='test_collect')
-        client = Client(database=backend)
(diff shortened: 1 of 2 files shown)


klauer:
Should a mongomock-less environment be part of the travis matrix?

ZLLentz:
You're getting hit by a style error.

Agree with Ken on mongo-mock free env for fully testing this, but it's not crucial because...

Shouldn't we just add a "pip install" section to our pcds-env scripts? There are going to be other packages like this and it's annoying to make little conda wrapper packages every time.

happi/tests/conftest.py
```diff
@@ -86,17 +86,19 @@ def mockjsonclient(device_info):
...
-        mock_mongo.return_value = mc
-        # Client
-        backend = MongoBackend(db='test_db',
-                               collection='test_collect')
-        client = Client(database=backend)
-        # Insert a single device
-        client.backend._collection.insert_one(device_info)
-        return client
-
+        if has_mongomock:
```
ZLLentz: nice

PR #92, opened 2019-03-12
Container

### LINKED ISSUE #83: Separate base Container and Device
There are a few LCLS specific things in `Device`. In reality we should have a base `Container` with the bare minimum globally useful keys, then have `Device` contain the more LCLS specific information i.e `stand`, `z`, `beamline`.

teddyrendahl:
## Description

For a while it has bothered me that base `happi.Device` makes a lot of assumptions about item you want to put in the database. I think location information should be required for physical devices but it is pretty unwieldy if you are just setting up a little simulation and want to fill a database with `ophyd.sim.SynAxis` objects.  

This PR creates a base `Container` object which has the bare minimum amount of `EntryInfo`. This is a good starting point for any external collaborators who want to use `happi` but don't plan on using any of our device types. One consequence of this is that we no longer enforce that `prefix` exists (see simulation) on `Container`. Therefore we need to retrain our database to use `name` as the unique enforceable key. This is still compatible with databases created with older versions of `happi` 

###### API Thoughts
In the docstrings I tried to replace all mentions of `Device` with `Container` but methods are still called `find_device` e.t.c. I like the new `Container` name, it creates a clear separation between the configuration information (`Container`) and the actual `ophyd.Device`. I think calling these the same thing is probably confusing for new folks.  

## Motivation and Context

Closes #83 

## How Has This Been Tested?

* Existing test suite passes, simply had to change a few places where we assumed what that `_id` tag was.
* Successfully loaded older happi database

TBD

### DIFF
--- a/happi/device.py
+++ b/happi/device.py
@@ -206,13 +206,13 @@ class InfoMeta(type):
         return clsobj
 
 
-class Container(metaclass=InfoMeta):
+class HappiItem(metaclass=InfoMeta):
     """
-    The base Container object
+    The smallest description of an object that can be entered in ``happi``
 
     The class does not need to be intialized with any specific piece of
     information except a name, but all of the attributes listed by
-    :attr:`Container.info_names` can be used to assign values to
+    :attr:`HappiItem.info_names` can be used to assign values to
     :class:`.EntryInfo` upon initialization.  Pieces of information that are
     deemed mandatory by the class must be filled in before the device is loaded
     into the database. See :attr:`Device.mandatory_info` to see which
@@ -246,7 +246,7 @@ class Container(metaclass=InfoMeta):
     -------
     .. code ::
 
-        d = Container(name = 'my_device',         #Alias name for device
+        d = HappiItem(name = 'my_device',         #Alias name for device
                       note  = 'Example',          #Piece of arbitrary metadata
                      )
     """
@@ -358,7 +358,7 @@ class Container(metaclass=InfoMeta):
         return (self.prefix, self.name) == (other.prefix, other.name)
 
 
-class Device(Container):
+class Device(HappiItem):
     """
     A Generic Device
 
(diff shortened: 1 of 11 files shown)


ZLLentz:
Biggest issue for me now is that I get really confused over the device/container verbiage when reading the codebase, but maybe that should be addressed separately.

I'd ultimately like the word `device` to be reserved for an object that actually "does something", rather than one that just stores information.

So we'd have `Container` and `PCDSContainer`, for example, instead of calling the second one `Device`, with appropriately awful API breaks as needed.

klauer:
I think the rename is important and definitely better to do sooner rather than later.  @ZLLentz and I came up with the following replacement class names:

* `Happi`
* `Descriptor`
* ~`Container`~
* ~`Device`~
* ~`Entry`~
* ~`Document`~
* ~`Row`~

We should brainstorm more...

teddyrendahl:
Renamed to `HappiItem`. Added to docs. Ready for final review and merge.

happi/device.py
```diff
@@ -382,3 +356,44 @@ def __repr__(self):
 
     def __eq__(self, other):
         return (self.prefix, self.name) == (other.prefix, other.name)
+
+
+class Device(HappiItem):
```
ZLLentz: We should pick a reasonable rename alias for this when we transition away from using this name

happi/device.py
```diff
@@ -382,3 +356,44 @@ def __repr__(self):
 
     def __eq__(self, other):
         return (self.prefix, self.name) == (other.prefix, other.name)
```
klauer: Not part of the PR, but this jumps out to me as being insufficient.
It should be something like:
`(type(self) is type(other)) and all((self.* == other.*))`

teddyrendahl: Good point. Let's make an issue

teddyrendahl:
Had to rebase so I could get the `happi/cli.py` file that is causing the `flake8` exception on Travis 🤕

happi/device.py
teddyrendahl: actually adressing this now, `current  __eq__ depends on all devices having a prefix

teddyrendahl:
A couple more issues. The old `__eq__` method required a `prefix` which is not on `HappiItem`. We now compare the full dump of the information that will be in the `happi` database. Not sure why we were not doing this before ...

The other thing is `happi` keeps a cache of the objects it loads. That cache was keyed on `prefix`... we now key on `name`. Finally, I noticed that if a device did not have metadata attached we would automatically return the cached object. This seems like the wrong behavior. If we can't guarantee the same device will be loaded we should side on reloading.

Worth another review @klauer @ZLLentz . Sorry for this dragging on.

PR #95, opened 2019-05-10
List choices for field

jsheppard95:
## Description
Added function to list choices for a given field that user can search for: `Client.list_choices`

## Motivation and Context
Will allow user to know what beamlines, prefixes, names, etc. will return results

## How Has This Been Tested?
`test_client.test_list_choices`

docstring of `Client.list_choices`
usage: Client.list_choices('beamline')
returns: list of beamlines the user can search for and will return results

### DIFF
--- a/happi/client.py
+++ b/happi/client.py
@@ -535,3 +535,24 @@ class Client:
         # If found nothing
         raise EnvironmentError("No happi configuration file found. "
                                "Check HAPPI_CFG.")
+
+    def list_choices(self, field):
+        """
+        List all choices for a given field
+
+        Parameters
+        ----------
+        field : string
+            search field to list all possible choices for
+            i.e 'beamline', 'name', 'z', etc.
+        """
+        field_choices = []
+        all_devs = self.all_devices
+        for dev in all_devs:
+            try:  # Want to ignore error if 'dev' doesn't have 'field'
+                choice = getattr(dev, field)
+                if choice not in field_choices:
+                    field_choices.append(choice)
+            except:
+                pass
+        return field_choices
--- a/happi/client.py
+++ b/happi/client.py
@@ -560,8 +560,7 @@ class Client:
         for dev in self.all_devices:
             try:  # Want to ignore error if 'dev' doesn't have 'field'
                 choice = getattr(dev, field)
-                if choice not in field_choices:
-                    field_choices.add(choice)
+                field_choices.add(choice)
             except AttributeError:
                 pass
         if len(field_choices) == 0:
--- a/happi/client.py
+++ b/happi/client.py
@@ -545,6 +545,16 @@ class Client:
         field : string
             search field to list all possible choices for
             i.e 'beamline', 'name', 'z', 'prefix', etc.
+
+        Raises
+        ------
+        SearchError
+            If no devices in the database have an entry for the given field
+
+        Returns
+        -------
+        field_choices : list
+            list of choices for a given field that are in the database
         """
         field_choices = []
         all_devs = self.all_devices
--- a/happi/client.py
+++ b/happi/client.py
@@ -544,7 +544,7 @@ class Client:
         ----------
         field : string
             search field to list all possible choices for
-            i.e 'beamline', 'name', 'z', etc.
+            i.e 'beamline', 'name', 'z', 'prefix', etc.
         """
         field_choices = []
         all_devs = self.all_devices
@@ -553,6 +553,6 @@ class Client:
                 choice = getattr(dev, field)
                 if choice not in field_choices:
                     field_choices.append(choice)
-            except:
-                pass
+            except AttributeError:
+                raise SearchError('No entries found with given field')
         return field_choices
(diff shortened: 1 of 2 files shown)


happi/client.py
```diff
@@ -535,3 +535,34 @@ def find_config():
...
+        SearchError
+            If no devices in the database have an entry for the given field
+
+        Returns
+        -------
+        field_choices : list
+            list of choices for a given field that are in the database
+        """
+        field_choices = []
+        all_devs = self.all_devices
```
teddyrendahl: You can just iterate over `all_devices`? `for dev in self.all_devices`

jsheppard95:
Bug found:
The added function does not work with all fields. will fail on:
`['_id', 'creation', 'data', 'last_edit', 'prefix_det', 'type']`
Likely due to the fact that these are not `EntryInfo` objects instantiated in `class Device`. Will need to either make these fields searchable or handle the error if passed to `list_choices`

happi/client.py
```diff
@@ -535,3 +535,34 @@ def find_config():
         # If found nothing
         raise EnvironmentError("No happi configuration file found. "
                                "Check HAPPI_CFG.")
+
+    def list_choices(self, field):
```
teddyrendahl: Trying to think of a name for this ... `list_choices` is kind of vague but I can't think of anything better. @ZLLentz @klauer?

happi/client.py
```diff
@@ -535,3 +535,34 @@ def find_config():
...
+        ------
+        SearchError
+            If no devices in the database have an entry for the given field
+
+        Returns
+        -------
+        field_choices : list
+            list of choices for a given field that are in the database
+        """
+        field_choices = []
```
teddyrendahl: Python has a data structure for this sort of thing called `Set`. It is essentially an unordered `List` that only allows unique values. I think it could be of some use to you here. 

https://www.w3schools.com/python/python_sets.asp

happi/client.py
```diff
@@ -535,3 +535,34 @@ def find_config():
...
+        all_devs = self.all_devices
+        for dev in all_devs:
+            try:  # Want to ignore error if 'dev' doesn't have 'field'
+                choice = getattr(dev, field)
+                if choice not in field_choices:
+                    field_choices.append(choice)
+            except AttributeError:
+                raise SearchError('No entries found with given field')
```
teddyrendahl: I don't think you want to raise `SearchError` this will stop on the first device that doesn't have the field. Imagine if you are searching for a custom field that is only on some containers. In my mind `SearchError` should only be raised if there are no containers with that field in the database.

happi/tests/test_client.py
```diff
@@ -226,3 +226,14 @@ def test_from_cfg(happi_cfg):
     client = Client.from_config()
     assert isinstance(client.backend, JSONBackend)
     assert client.backend.path == 'db.json'
+
+
+def test_list_choices(mockjsonclient):
```
teddyrendahl: You can probably just use the `happi_client` fixture instead of just the `mockjsonclient`

happi/client.py
klauer: Maybe:
```suggestion
    def choices_for_field(self, field):
```

teddyrendahl: Think I like it better

ZLLentz: `choices_for_field`
`choices`
`field_choices`
`options_for_field`
`options`
`field_options`

jsheppard95: Changed in 2191d86

happi/client.py
jsheppard95: Changed in 2191d86

happi/tests/test_client.py
jsheppard95: Changed in 2191d86

happi/client.py
jsheppard95: This I'm still stuck on. I changed it so that I only raise `SearchError` if `field_choices` (which starts as an empty set) is still empty after iterating through all devices. This works as one can now do:
`client.choices_for_field('n_filters')`
and returns:
`{6, 10}`
But I still seem to be unable to show the choices for some fields, specifically `type` or `_id`. Still trying to sort out why this is and how to fix it.

teddyrendahl: Those don't get set as attributes on the device, but will only show up if you look at `post()`. In general, users should not be searching these fields.

happi/client.py
```diff
@@ -535,3 +535,35 @@ def find_config():
...
+        Returns
+        -------
+        field_choices : list
+            list of choices for a given field that are in the database
+        """
+        field_choices = set()
+        for dev in self.all_devices:
+            try:  # Want to ignore error if 'dev' doesn't have 'field'
+                choice = getattr(dev, field)
+                if choice not in field_choices:
```
teddyrendahl: Do you need this `if` anymore? Think `Set` handles this for you.

jsheppard95: I most certainly do not. Changed in a89f330

PR #99, opened 2020-02-24
ENH: add cli commands, update required fields, fix issues

ZLLentz:
## Description

- Add an `add` command for cli, e.g. `happi add` to start an interactive device adder
- Add an `edit` command for cli, e.g. `happi edit im3l0 location=750 prefix=IM3L0:PPM`
- Change `search` command syntax to be simpler (more like `edit`)
- Add a `load` command for cli, e.g. `happi load im3l0 im1l1` -> `IPython` session
plus other changes made in dev to "get it working"

## Motivation and Context

I got really upset trying to remember how to do all of these in the shell
- `add` will help us populate the database
- `edit` will help us fix mistakes
- `load` will help us debug classes and see if the entries are correct

## How Has This Been Tested?

Only live tests so far, should not be merged yet

This will need to be added to the docs cli page

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -11,7 +11,6 @@ from IPython import start_ipython
 import coloredlogs
 
 import happi
-from .errors import SearchError
 
 logger = logging.getLogger(__name__)
 
@@ -31,8 +30,8 @@ subparsers = parser.add_subparsers(help='Subparsers to search, add, edit',
                                    dest='cmd')
 parser_search = subparsers.add_parser('search', help='Search the happi '
                                       'database')
-parser_search.add_argument('search_criteria', nargs=argparse.REMAINDER,
-                           help='search criteria: FIELD VALUE')
+parser_search.add_argument('search_criteria', nargs='+',
+                           help='search criteria: FIELD=VALUE')
 parser_add = subparsers.add_parser('add', help='Add new entries')
 parser_edit = subparsers.add_parser('edit', help='Change existing entry')
 parser_edit.add_argument('name', help='Device to edit')
@@ -72,18 +71,10 @@ def happi_cli(args):
     if args.cmd == 'search':
         logger.debug("We're in the search block")
 
-        # Ensure we have an even number of search elements
-        # Should always have key:value pairs
-        if len(args.search_criteria) % 2 != 0:
-            raise SearchError('Search criteria should be given as key '
-                              'value pair\ni.e:\n'
-                              'happi search beamline MFX stand DG1')
-
         # Get search criteria into dictionary for use by client
         client_args = {}
-        for i in range(0, len(args.search_criteria), 2):
-            criteria = args.search_criteria[i]
-            value = args.search_criteria[i+1]
+        for user_arg in args.search_criteria:
+            criteria, value = user_arg.split('=')
             if value.replace('.', '').isnumeric():
                 logger.debug('Changed %s to float', value)
                 value = float(value)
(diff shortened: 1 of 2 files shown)


happi/device.py
```diff
@@ -394,7 +394,7 @@ class Device(HappiItem):
     args = copy.copy(HappiItem.args)
     args.default = ['{{prefix}}']
     kwargs = copy.copy(HappiItem.kwargs)
-    args.default = {'name': '{{name}}'}
+    kwargs.default = {'name': '{{name}}'}
```
ZLLentz: FYI happi has been adding devices with the wrong default args and kwargs for at least a year

ZLLentz:
I think I've done everything I want to do here. Will need to revisit to add/adjust docs and tests.

happi/cli.py
```diff
@@ -59,18 +71,10 @@ def happi_cli(args):
...
-                              'value pair\ni.e:\n'
-                              'happi search beamline MFX stand DG1')
-
         # Get search criteria into dictionary for use by client
         client_args = {}
-        for i in range(0, len(args.search_criteria), 2):
-            criteria = args.search_criteria[i]
-            value = args.search_criteria[i+1]
+        for user_arg in args.search_criteria:
+            criteria, value = user_arg.split('=')
```
klauer:
```suggestion
            criteria, value = user_arg.split('=', 1)
```

happi/cli.py
```diff
@@ -83,6 +87,63 @@ def happi_cli(args):
...
+        if 'y' in ok:
+            logger.info('Adding device')
+            device.save()
+        else:
+            logger.info('Aborting')
+    elif args.cmd == 'edit':
+        logger.debug('Starting edit block')
+        device = client.find_device(name=args.name)
+        for edit in args.edits:
+            field, value = edit.split('=')
```
klauer:
```suggestion
            field, value = edit.split('=', 1)
```

ZLLentz:
Also: it's very likely that we will find that some of the edit fields don't work through this cli and will need some special string handling. I will tackle this when I add tests.

I'm not in a rush to merge this, I'm happy to let it sit as we run out of dev.

happi/cli.py
ZLLentz: To clarify, is this for inputs like `name=Ilikethe=sign`?

klauer: Yes. However unlikely it is that there will be `=` in `value`, it's easier to just build this in now.

klauer:
Should also check to see if `'='` is in the string for searching, as I just ran into:

```
(dev) [klauer@psdev02  happi]$ happi search K-Mono
Traceback (most recent call last):
  File "/reg/g/pcds/pyps/conda/py36/envs/dev/bin/happi", line 11, in <module>
    load_entry_point('happi==1.3.0', 'console_scripts', 'happi')()
  File "/reg/g/pcds/pyps/apps/dev/pythonpath/happi/cli.py", line 163, in main
    happi_cli(sys.argv[1:])
  File "/reg/g/pcds/pyps/apps/dev/pythonpath/happi/cli.py", line 79, in happi_cli
    criteria, value = user_arg.split('=', 1)
ValueError: not enough values to unpack (expected 2, got 1)
```

ZLLentz:
tbh `happi search <name>` should just be an alternate syntax for searching by name

klauer:
Definitely!

ZLLentz:
it has been done

ZLLentz:
Merging this as discussed on slack chat to avoid PR getting stale. We should revisit to add tests and clean up bugs before the next release.

Issue #106, opened 2020-03-10
Discussion about Device and Mandatory Fields


hhslepicka:
## Expected Behavior
Since other labs are looking into Happi, should we make Device be completely flexible with additional metadata and create the PCDSDevice or some other name that will have mandatory fields such as `functional_group` and `location_group` ?

## Current Behavior
Device requires `functional_group` and `location_group` along with some other information that is a bit LCLS specific.

## Possible Solution
Create a new type of container that is the PCDS basic Device and make Device very relaxed.

hhslepicka:
From SLACK talk:
- Zach says: "My answer is yes"
- Ken says: 👍

ZLLentz:
I think we should use this opportunity to change our semantics too. These should never have been called `Device` because of the conflict with `ophyd`. I think `HappiItem` should be the super basic entry, then if you want an ophyd device entry it should be `DeviceItem`, and then more specific should be `SlitsItem`, etc. if we want these.

I would also advocate for pulling all SLAC-specific assumptions out of `happi` and into e.g. `happi_lcls`

ZLLentz:
There would also need to be a way to register your site-specific classes with `happi`

klauer:
Not that I'm against it necessarily, I think pulling things out into a separate package for PCDS/LCLS is probably slightly overkill at this point.

A minimum viable thing to make this useful for others and still keep PCDS stuff our priority would be refactoring as @hhslepicka suggests, I think.

hhslepicka:
Discussion from #112:

> The container discovery will need most of the refactor.(https://github.com/pcdshub/happi/blob/master/happi/client.py#L44-L52)
> 
> Here is one idea:
> Make containers be converted from a python file to a namespace package (https://packaging.python.org/guides/packaging-namespace-packages/#creating-a-namespace-package) with a registry inside of it.
> Facilities that wants to add new containers can create a package which will share this namespace and be discovered by Happi when running.
> 
> Another idea (by @klauer):
> Investigate the usage of entry_points so that we don't need to enforce a package architecture into projects that want to extend the containers for Happi.

PR #112, opened 2020-03-16
REF: remove metaclass for HappiItem + allow for easy dict interface

klauer:
I think this will be a good spot to jump on #106 for other lab's use - want to give that a go @hhslepicka?

### DIFF
--- a/happi/device.py
+++ b/happi/device.py
@@ -1,3 +1,4 @@
+import collections
 import copy
 import logging
 import re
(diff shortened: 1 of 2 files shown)


hhslepicka:
@klauer sure, I just think we need to discuss some key points before jumping to the implementation to make sure we address the comments by @ZLLentz at #106 .

The container discovery will need most of the refactor. (https://github.com/pcdshub/happi/blob/master/happi/client.py#L44-L52)

Here is one idea:
Make `containers` be converted from a python file to a namespace package (https://packaging.python.org/guides/packaging-namespace-packages/#creating-a-namespace-package) with a registry inside of it.
Facilities that wants to add new containers can create a package which will share this namespace and be discovered by Happi when running.

klauer:
Cool - I think we're on the same page. @hhslepicka mentioned in a chat he'll investigate the alternatives and come to a conclusion.

Probably should move the discussion back to the issue, as the basis this PR provides is somewhat tangential to what I suggested above...

hhslepicka:
Let's merge this one in. I copied my comment to the issue. I will open a separated PR so we can discuss implementation details.

PR #115, opened 2020-03-16
REF: Move from metaclass Backend to base class.

hhslepicka:
## Description
Switching to a base class as in this case of the backend there is no particular reason (as far as I could tell) for it to be a metaclass.

## How Has This Been Tested?
Tests are passing.

### DIFF
--- a/happi/backends/core.py
+++ b/happi/backends/core.py
@@ -1,24 +1,22 @@
 """
-Abstract backend database options
+Base backend database options
 """
-import abc
 import logging
 
 logger = logging.getLogger(__name__)
 
 
-class Backend(abc.ABCMeta):
+class _Backend:
     """
-    Abstract interface for backend database
+    Base class for backend database
     """
-    @abc.abstractproperty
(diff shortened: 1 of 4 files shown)


klauer:
flake8 failing :(

happi/backends/core.py
```diff
@@ -1,24 +1,22 @@
...
-class Backend(abc.ABCMeta):
+class _Backend:
     """
-    Abstract interface for backend database
+    Base class for backend database
     """
-    @abc.abstractproperty
-    def devices(self):
+
+    def all_devices(self):
```
klauer: Let's keep this a `@property` otherwise it's callable now instead

hhslepicka: It was deleted by accident. Addressed via https://github.com/pcdshub/happi/pull/115/commits/e44f98f7a5daeac13ef52c1a07c66d907260eb2e

ZLLentz:
This looks good to me as well a day later

PR #116, opened 2020-03-17
CI: tweak CI and add python 3.8

klauer:
No way this'll work on the first try. More tomorrow.


klauer:
Maybe not the recommended setup, but it's somewhere closer...

PR #117, opened 2020-03-17
ENH: allow copying of devices

### LINKED ISSUE #1: Copy method for Device
Device Container class should have a copy special method

klauer:
Closes #1

Also, my first attempt at using the new `gh` CLI to list issues + create a PR.
Apologies if this doesn't turn out as I expect...

### DIFF
--- a/happi/device.py
+++ b/happi/device.py
@@ -273,12 +273,10 @@ class HappiItem(_HappiItemBase, collections.abc.Mapping):
             else:
                 setattr(self, info.key, info.default)
         # Handle additional information
+        self.extraneous = kwargs
         if kwargs:
             logger.debug('Additional information for %s was defined %s',
-                         self.name, ', '.join(kwargs.keys()))
-            self.extraneous = kwargs
-        else:
-            self.extraneous = {}
+                         self.name, ', '.join(self.extraneous))
 
     @property
     def info_names(self):
@@ -347,6 +345,12 @@ class HappiItem(_HappiItemBase, collections.abc.Mapping):
 
     def __len__(self):
         return len(self.post())
+
+    def __copy__(self):
+        return type(self)(**dict(self))
+
+    def __deepcopy__(self, memo):
+        return self.__copy__()
 
     @property
     def screen(self):
--- a/happi/device.py
+++ b/happi/device.py
@@ -347,10 +347,11 @@ class HappiItem(_HappiItemBase, collections.abc.Mapping):
         return len(self.post())
 
     def __copy__(self):
-        return type(self)(**dict(self))
+        return self.__deepcopy__({})
 
     def __deepcopy__(self, memo):
-        return self.__copy__()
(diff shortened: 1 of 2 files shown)


ZLLentz:
Do we have a junk build of psdm_qs_cli on our dev channel? The CI failures don't seem to be related to this PR.

klauer:
It appears there is some bad package there... Though not a solution per-se, I'm hoping to side-step that issue soon with psdm* + krtc on conda-forge.

ZLLentz:
I think this is something that can happen when you upload the same version of a package multiple times to the same channel, as we do in the travis for that library

happi/device.py
```diff
@@ -348,6 +346,12 @@ def __iter__(self):
     def __len__(self):
         return len(self.post())
 
+    def __copy__(self):
+        return type(self)(**dict(self))
+
+    def __deepcopy__(self, memo):
+        return self.__copy__()
```
ZLLentz: I'm rusty on this, are you overriding some default behavior in the `__deepcopy__`? Or is there no default behavior for either method using the other and we must always supply both a `__copy__` and a `__deepcopy__` to enable copying?

klauer: I don't believe `__deepcopy__` calls `__copy__` by default, so it's up to us to implement.

It's possible this 'deep copy = shallow copy' as I've implemented it may come back to bite us for mutable objects contained within the device. We could wait to figure that out or consider this "good enough for now" - up to you guys.

ZLLentz: The alternative is to call deepcopy recursively on the dict items, right? I don't think it matters that much, but maybe it's better to have no deepcopy at all than one that lies to us?

klauer: Quite reasonable. It was dumb of me not to implement it properly

hhslepicka: not dumb... just what you could get as good enough for now and that makes sense! 😉

klauer: Thanks, guys. That's what these reviews are about after all! Thinking about it a bit more, maybe it's not as bad as I thought.

How's 39edf8c look? It flips this around such that `__copy__` and `__deepcopy__` always return a deep copy with the help of the `copy` module. 

A shallow copy when you ask for a deepcopy is without question bad. A deep copy is safer in all regards, resulting in less confusion for users.

hhslepicka: So now copy returns a deepcopy as well? At #39edf8c I am also confused because `memo` is not used... should you add the copy to `memo`? I am not familiar with implementing __deepcopy__ sorry if dumb question.

ZLLentz: I'm also not savvy on the `memo` argument but this implementation looks good to me.

klauer: `memo` is for keeping track of making the deepcopy, so you can stash stuff that newly-created sub-objects might need to be copied.

It should be fine as-is, since we just have a [potentially nested] dictionary of built-in types in the form of the device information.

PR #118, opened 2020-03-17
MNT: Multi backend

### LINKED ISSUE #24: Support multiple configurations
Allow a user to have multiple backends loaded. This allows for shared configuration files for common areas. The database should be smart enough to know which configuration file to add the device.

The only really tricky part is allowing the user to use different backends and still add new devices

klauer:
I set out to further understand happi internals and ~close~ #24, but in writing the commits here, I think there may not be anything to fix.

The couple things here add a bit of clarity and also allow access to all loaded backends - which could potentially be an entrypoints hook at some point, I suppose. So, this doesn't have to be merged 🤷‍♂ 

I thought we were going to have a single database for beamline devices, at least. Other uses of happi (which do not yet exist) should load a different configuration that only they are aware of?

Perhaps @ZLLentz could further clarify and decide whether or not we can ~close~ #24.

### DIFF
--- a/happi/backends/__init__.py
+++ b/happi/backends/__init__.py
@@ -1,24 +1,42 @@
-__all__ = ['backend']
+__all__ = ['backend', 'BACKENDS', 'DEFAULT_BACKEND']
 import os
-# Check to see if the user has specified a specific backend
-# as an environment variable. Import this as the standard
-# backend for other places in the module. A user can always
-# override this by explicitly importing the backend
-_backend = os.environ.get("HAPPI_BACKEND", 'json').lower()
-
 
 def _get_backend(backend):
     if backend == 'mongodb':
         from .mongo_db import MongoBackend
         return MongoBackend
-    elif backend == 'json':
+    if backend == 'json':
         from .json_db import JSONBackend
         return JSONBackend
-    else:
-        raise ImportError("Improper specification of happi backend."
-                          "Check `$HAPPI_BACKEND` environment variable.")
+    raise ValueError(f'Unknown backend {backend!r}')
 
 
-backend = _get_backend(_backend)
+def _get_backends():
+    # A hook for entrypoints or something similar later
+    backends = {}
+    try:
+        backends['json'] = _get_backend('json')
+    except ImportError as ex:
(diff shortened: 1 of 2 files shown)


happi/backends/__init__.py
```diff
@@ -1,24 +1,42 @@
...
+    try:
+        backends['json'] = _get_backend('json')
+    except ImportError as ex:
+        logger.warning('JSON backend unavailable: %s', ex)
+
+    try:
+        backends['mongodb'] = _get_backend('mongodb')
+    except ImportError as ex:
+        logger.warning('MongoDB backend unavailable: %s', ex)
+
```
hhslepicka: We are missing here probably the Questionnaire backend?

ZLLentz:
Based on the line "shared configuration files for common areas", I'm assuming the intention of #24 was to allow us to import from multiple files, e.g. we could have `lfe_devices.json` and maintain it separately from `xrt_devices.json`, and have some way to open a session where when we save a particular device we know which file to modify.

klauer:
I see, @ZLLentz. Based on what you've said, I think it's probably closable, as that is certainly supported.

hhslepicka:
maybe on happi the change would be to allow a client to have multiple databases but in this case one would need to specify a database as the writable one in some way.

hhslepicka:
@klauer not quite supported due to the fact that you would need to do a search on every client instance...

If Client accepted more than one database a call to `search` would automatically look at all database instances assigned to that client.

ZLLentz:
E.g., have one client with multiple databases instead of many clients with one database each?

klauer:
I suppose we could have a "client proxy"-like object that combines multiple clients and fans out such requests? Still seems weird to me to have that in happi instead of on the application end.

ZLLentz:
If it's something that multiple clients will eventually implement in essentially the same way, it's probably worth including in the library

klauer:
Fair enough, I suppose. Who is the first one who will use happi in the way you describe?

ZLLentz:
Probably us if we get fed up enough with 1 file to want to split but not fed up enough with json to set up a mongodb

Issue #98, opened 2020-01-07
Support fuzzy search


klauer:
LUCID will be using fuzzywuzzy (https://github.com/pcdshub/lucid/pull/10) so we may as well use the dependency here, too.

klauer:
I'm not crazy about fuzzywuzzy (or how lucid implemented it).
Consider investigating more on how FZF does matching as it more often than not returns what I expect:

https://github.com/junegunn/fzf/blob/master/src/algo/algo.go

klauer:
To investigate: https://github.com/taleinat/fuzzysearch

(available on PyPI, not conda-forge)

klauer:
Shell-level fuzzy completion may even be a good first step here. 

This little incantation may even be worthwhile for us:
```
happi search '*' |grep '| name' | awk -F '|' '{print $3}'  | sed 's/ //g' | fzf
```

To see it in action: https://asciinema.org/a/358799

PR #120, opened 2020-03-18
ENH: happi.client.Client.search() refactor

### LINKED ISSUE #4: Search on Regex
Be able to search the database on Regex patterns

klauer:
## Description
Working on getting full-match find alternatives such as:
* Find by regex
* Fuzzy find

## Motivation and Context
Closes #4 
Hopes to address #98 eventually

Docstrings, not docs yet

## API change
I expect that different backends could support different search methods, requiring only `find` to be fully supported. 

I'm definitely open to suggestions as to how this should look.

### DIFF
--- a/happi/client.py
+++ b/happi/client.py
@@ -32,11 +32,31 @@ def _looks_like_database(obj):
 
 
 class SearchResult(collections.abc.Mapping):
-    def __init__(self, client, metadata, *, device=None):
+    '''
+    A single search result from ``Client.search``
+
+    This result can be keyed for metadata as in::
+        result['name']
+
+    The HappiItem can be readily retrieved::
+        result.device
+
+    Or the object may be instantiated::
+        result.get()
+
+    Attributes
+    ----------
+    device : happi.HappiItem
+        The container
+    metadata : dict
+        The HappiItem metadata
+    '''
+
+    def __init__(self, client, device):
         self._device = device
         self._instantiated = None
         self.client = client
-        self.metadata = metadata
+        self.metadata = device.post()
 
     @property
     def device(self):
@@ -53,10 +73,6 @@ class SearchResult(collections.abc.Mapping):
                 threaded=threaded
             )
         return self._instantiated
-
-    def post(self):
-        'Return the full happi.Device post metadata'
-        return self.device.post()
 
     def __getitem__(self, item):
         return self.metadata[item]
@@ -339,7 +355,12 @@ class Client(collections.abc.Mapping):
 
     def __getitem__(self, key):
         'Get a device ID'
-        return SearchResult(client=self, metadata=self.backend.get_by_id(key))
+        try:
+            device = self.find_device(**self.backend.get_by_id(key))
+        except Exception as ex:
+            raise KeyError(key) from ex
+
+        return SearchResult(client=self, device=device)
 
     def __iter__(self):
         for info in self.backend.find({}):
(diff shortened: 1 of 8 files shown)


klauer:
If we like the API here, I'll go back to the docs.

This splits up searches into several methods on the backend + client for flexibility/clarity. 
Backends also implement these routines as generators, removing the confusing `multiple` kwarg from `find()`. Implemented both backends, which gives questionnaire support by the nature of it being a JSON backend subclass.

```
client.search(key=value)  # same as before
client.search_range('z', start=0, stop=100, key=value, **other_stuff)  # before 'z' was implicit in `search()`
client.search_regex(key=r'reg\dex', key2=r'a[123]c')  #  new, flags default to ignorecase
```

`client.search_fuzzy` is not there yet, and will fall to another PR.

Also adds:
```
client[name] -> doc
```
I'm thinking this instead should return a device, but would like some input.

hhslepicka:
I really like how it is coming together. API-wise it is fantastic.
Should we add the segmentation of results at all search operations?

About returning the device, is this idea just for the `client[name]->doc`  or all search opeartions?
If only for the `get_by_id` I agree that it would be okay to return the device.
If the idea is to apply that to all search then I am against it as in most cases we should delay creating the device until the last moment (UI mainly but can also be useful for hutch-python). Sometimes we just want to render the list of devices and the `HappiItem` is enough for that... Maybe that could be an option but I think that it will bring in complications to do it in a separated thread etc, right?

ZLLentz:
Looks clean to me, I like it. Similar thoughts as Hugo RE: doc vs device on key... Are there any situations where:
1. you know the database key
2. you want to check the metadata, but you want to avoid making an ophyd device

If so, then it should stay as is. If not... it'd be a cool shortcut to just do `client[im2l0].get()`

klauer:
Hmm, I think there may be some confusion as to "happi Device [container]" and "ophyd Device" as well. I made this metadata wrapper in the past 2 commits. I wonder, though, if I had clarified above as to which "Device" I was referring to what your opinions would have been.

And what we're going to do about disambiguation in the future...

happi/client.py
```diff
@@ -290,25 +335,69 @@ def all_devices(self):
...
+    def __len__(self):
+        return len(self.all_devices)
 
-    def search(self, start=0., end=None, as_dict=False, **kwargs):
+    def _get_search_results(self, items, *, wrap_cls=None):
+        '''
+        Return search results to the user, optionally wrapping with a class
+        '''
+        wrap_cls = wrap_cls or self._results_wrap_class
+        return [wrap_cls(client=self, metadata=info,
```
hhslepicka: Isn't this defeating the `yield` of the results?

klauer: It removes the benefit of having the generator in the backend, yes. Should we have access to the generators in the user-facing API?

klauer:
* Add alternative `search` methods on the client which act as generators (name?)
* Make the LCLS-specific device stuff
* Change naming of `device` -> `item` (or `container`? or `description`? or `metadata`?) in many places (name TBD)
* Entry-points for containers
* Documentation rewrite for all of the above

I think this PR is probably good as-is to serve as a basis for the other tasks. If you're OK with the changes, shall we merge and continue on the rest? @ZLLentz @hhslepicka

hhslepicka:
@klauer I agree, let's merge this PR and move the comments that you added into an issue for follow up work.

PR #121, opened 2020-03-18
BLD/CLN: Add pre-commit hook and clean the repo

klauer:
Let's give `pre-commit` a try. What do you think @hhslepicka @ZLLentz?

```bash
$ brew install pre-commit # or install via pip
$ cd happi
$ pre-commit install          # install for this repo based on the config
$ pre-commit run --all-files  # run on everything
$ pre-commit run              # run on staged
$ git commit -am "test"       # run pre-commit and - if successful - commit
```


ZryletTC:
Curious why it's seeing the wrong hashes 🤔

hhslepicka:
@ZryletTC where?

klauer:
@ZryletTC ugh, CI is all kinds of busted at the moment. 

In this case it may have been me trying out the new `gh` command-line tool (it's pretty neat, check it out: https://cli.github.com/) that screwed things up a bit further since it defaulted to pushing to `pcdshub:pre-commit` first instead of `klauer:pre-commit`. I removed that branch and now it looks confused? Hmmm...

ZLLentz:
I like this a lot. The summary is: have a python environment with the pre-commit thing, set it up on your repo when you clone it? Does something need to be gitignored or does it stash magic in the `.git` folder?

klauer:
Summary is as follows:
* Make a `pre-commit-config.yaml` in a repository
* Optionally version control that configuration
* Run `pre-commit install` in that repository
* Commit away
* Relevant info is stashed in `.git` and there's no need for `.gitignore`

I think for us:
* We add `pre-commit-config.yaml` to our repos
* Eventually we come up with our own `pcds-commit-hooks` repo
* The only pain point would be updating `pre-commit-config.yaml` on each repo we have, similar to different travis configurations

ZryletTC:
> @ZryletTC where?
In the Travis CI pr build, conda was stopping because it was attempting to compare the hashes between these two files:
Screen Shot 2020-03-18 at 12 20 47 PM

hhslepicka:
@ZryletTC it probably was a glitch at the anaconda server. Usually you can upload the same package many times and it won't fail. Someone already deleted the repeated package from the pcds-dev channel. Let's keep the eye open on that. Great finding.

hhslepicka:
I installed it here and it looks great! I think we should add that to all of our projects. Big 👍

PR #122, opened 2020-03-19
DEP: Rename device

ZLLentz:
## Description

- Deprecate `Device` and `device.py`
- Move `HappiItem` to `item.py`
- Establish `OphydItem` and `LCLSItem`
- Cut unused fields while we're at it

## Motivation and Context

- `Device` is too confusing with the conflicting names in `ophyd`
- We don't use most of these fields. They can be added later when a need arises.

## How Has This Been Tested?

Not yet, looking for feedback. Old tests still pass, now with warnings.

Not yet, looking for feedback.


happi/item.py
```diff
@@ -0,0 +1,392 @@
...
+logger = logging.getLogger(__name__)
+
+# For back-compat to <py3.7
+try:
+    from re import Pattern
+except ImportError:
+    from re import _pattern_type as Pattern
+
+
+class EntryInfo(object):
```
klauer:
```suggestion
class EntryInfo:
```

happi/item.py
```diff
@@ -0,0 +1,392 @@
...
+                                                               self.default)
+
+    def __copy__(self):
+        return EntryInfo(doc=self.doc, optional=self.optional,
+                         enforce=self.enforce, default=self.default)
+
+
+class _HappiItemBase:
+    def __init_subclass__(cls, **kargs):
+        # These attributes are used by device so can not be overwritten
```
klauer:
```suggestion
        # These attributes are not to be overwritten by subclasses
```

happi/item.py
```diff
@@ -0,0 +1,392 @@
...
+            else:
+                setattr(self, info.key, info.default)
+        # Handle additional information
+        self.extraneous = kwargs
+        if kwargs:
+            logger.debug('Additional information for %s was defined %s',
+                         self.name, ', '.join(self.extraneous))
+
+    @property
+    def info_names(self):
```
klauer: We should move these into the `_HappiItem` `__init_subclass__`, as `info_names`, `mandatory_info` does not change.

ZLLentz: Done

ZLLentz:
Next thing to do is to go through all of our methods that include `device` in the name, deprecate them, and give an alternate method with `item` in the name

ZLLentz:
I'm also going to be going through internal source code and swapping `device` for `item` wherever appropriate to boost readability

hhslepicka:
That is a great idea, @ZLLentz

happi/device.py
```diff
@@ -415,14 +52,28 @@ class Device(HappiItem):
...
                                     self.__class__.__name__,
                                     self.name,
                                     self.prefix,
                                     self.z)
+
+    @property
+    def screen(self):
+        warnings.warn("The 'screen' keyword is no longer used in Happi as it "
+                      "lacks specificity. Use one of detailed_screen, "
+                      "embedded_screen, or engineering screen instead")
```
hhslepicka:
```suggestion
                      "embedded_screen, or engineering_screen instead")
```

happi/device.py
```diff
@@ -415,14 +52,28 @@ class Device(HappiItem):
...
+    parent = EntryInfo('If the device is a component of another, '
+                       'enter the name', enforce=str)
     args = copy.copy(HappiItem.args)
     args.default = ['{{prefix}}']
     kwargs = copy.copy(HappiItem.kwargs)
     kwargs.default = {'name': '{{name}}'}
 
+    def __init__(self, *args, **kwargs):
+        warnings.warn("happi.device.Device is deprecated. Please use "
+                      "OphydItem or LCLSItem.")
```
hhslepicka:
```suggestion
                      "OphydItem or LCLSItem.", DeprecationWarning)
```

PR #126, opened 2020-03-20
ENH: Containers via entrypoints

### LINKED ISSUE #106: Discussion about Device and Mandatory Fields

hhslepicka:
## Description
As of now in order to add a new container we need to modify Happi source code and add it at the `happi.containers` module so it is found by the client.
Via discussion at #106 and many talks with @ZLLentz and @klauer we decided to experiment with the usage of entry_points.
The LCLS specific containers and item were moved to `pcdsdevices` under the following PR: https://github.com/pcdshub/pcdsdevices/pull/402

## Motivation and Context

Merging this will contribute to close #106 

## How Has This Been Tested?

Tested locally

Still need to be done. Will work on that after code comments

```python
In [1]: import happi
Questionnaire backend unavailable: No module named 'psdm_qs_cli'

In [2]: happi.containers.registry
Out[2]:
{pcdsdevices.happi.containers.Acromag,
 pcdsdevices.happi.containers.AreaDetector,
 pcdsdevices.happi.containers.Attenuator,
 pcdsdevices.happi.containers.BeamControl,
 pcdsdevices.happi.containers.Diagnostic,
 pcdsdevices.happi.containers.GateValve,
 pcdsdevices.happi.containers.IPM,
 pcdsdevices.happi.containers.LCLSItem,
 pcdsdevices.happi.containers.LODCM,
 pcdsdevices.happi.containers.Motor,
 pcdsdevices.happi.containers.MovableStand,
 pcdsdevices.happi.containers.OffsetMirror,
 pcdsdevices.happi.containers.PIM,
 pcdsdevices.happi.containers.PulsePicker,
 pcdsdevices.happi.containers.Slits,
 pcdsdevices.happi.containers.Stopper,
 pcdsdevices.happi.containers.Trigger,
 pcdsdevices.happi.containers.Vacuum}

In [3]: happi.containers.Acromag
Out[3]: pcdsdevices.happi.containers.Acromag

In [4]: dir(happi.containers)
Out[4]:
['Acromag',
 'AreaDetector',
 'Attenuator',
 'BeamControl',
 'Diagnostic',
 'GateValve',
 'HappiItem',
 'IPM',
 'LCLSItem',
 'LODCM',
 'Motor',
 'MovableStand',
 'OffsetMirror',
 'PIM',
 'PulsePicker',
 'Slits',
 'Stopper',
 'Trigger',
 'Vacuum',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__spec__',
 '_entries',
 'entry',
 'entrypoints',
 'inspect',
 'logger',
 'logging',
 'obj',
 'registry']
```


happi/containers.py
```diff
@@ -1,330 +1,30 @@
...
+
+registry = set()
+
+for entry in _entries:
+    try:
+        obj = entry.load()
+    except Exception:
+        logger.exception("Failed to load happi.containers entry: %s", entry)
+        continue
+    if inspect.isclass(obj) and HappiItem in obj.mro():
```
klauer:
```suggestion
    if inspect.isclass(obj) and issubclass(obj, HappiItem):
```

happi/containers.py
```diff
@@ -1,330 +1,30 @@
...
+        registry.add(obj)
+    elif inspect.ismodule(obj):
+        registry.update([var for _, var in obj.__dict__.items()
+                         # Only classes
+                         if inspect.isclass(var)
+                         # Must inherit from HappiItem at some point
+                         and HappiItem in var.mro()
```
klauer:
```suggestion
                         and issubclass(var, HappiItem)
```

happi/containers.py
```diff
@@ -1,330 +1,30 @@
...
+                         # Avoid happi internal classes due to imports
+                         and not var.__module__.startswith('happi.')])
+
+        locals().update({c.__name__: c for c in registry})
```
klauer: Indentation right here (seems like this should happen at the end after the registry is built)?

I think you're trying to get `happi.containers.<classname>` working: `globals()` seems more correct but may do the same here. We're also risking name collisions in the namespace as this strips off the `__module__` prefix, hmm...

happi/containers.py
```diff
@@ -1,330 +1,30 @@
...
+import inspect
+import entrypoints
+
+from .item import HappiItem
+
+logger = logging.getLogger(__name__)
+
+_entries = entrypoints.get_group_all('happi.containers')
+
+registry = set()
```
klauer: I wonder if the registry should be keyed as follows: `registry['(module).(classname)'] = obj`?

happi/containers.py
hhslepicka: Solved with https://github.com/pcdshub/happi/pull/126/commits/5c39ffbb8d283184a295b6080631c60f033c681e. Sorry that I forgot to push before opening the PR.

happi/containers.py
hhslepicka: Solved with https://github.com/pcdshub/happi/pull/126/commits/5c39ffbb8d283184a295b6080631c60f033c681e. Sorry that I forgot to push before opening the PR.

happi/containers.py
hhslepicka: Indentation solved with https://github.com/pcdshub/happi/pull/126/commits/5c39ffbb8d283184a295b6080631c60f033c681e. Sorry that I forgot to push before opening the PR.

hhslepicka: About the globals()... I'm not sure.

hhslepicka: ok, now I understand... that is correct, we inject the class name into the `happi.containers` namespace it would fail if we have 2 containers with the same name but different full qualified name.

happi/containers.py
hhslepicka: What about the `name` from the entry_points? This way we can make it uniform to always be the package that originated it and we guarantee unique names per package?
E.g.:
`pcdsdevices.Acromag` as the container and key at registry.

hhslepicka:
I need to fix the tests and I am finding some more issues... it will take a while, folks! 😞

happi/containers.py
klauer: Hmm, maybe... but what about multi-level imports?

Would (e.g.) `pcdsdevices.motors.Acromag` become `pcdsdevices.Acromag`?

happi/containers.py
```diff
@@ -1,330 +1,27 @@
...
+import inspect
+import entrypoints
+
+from .item import HappiItem
+
+logger = logging.getLogger(__name__)
+
+_entries = entrypoints.get_group_all('happi.containers')
+
+registry = {}
```
klauer: Let's move the `registry` generation code inside a function, otherwise things like `name`, `entry`, `obj`, etc may leak such that `from happi.containers import entry` would work.

happi/containers.py
```diff
@@ -1,330 +1,27 @@
...
+for entry in _entries:
+    try:
+        name = entry.name
+        obj = entry.load()
+    except Exception:
+        logger.exception("Failed to load happi.containers entry: %s", entry)
+        continue
+    if inspect.isclass(obj) and issubclass(obj, HappiItem):
+        registry[f"{name}.{obj.__name__}"] = obj
+    elif inspect.ismodule(obj):
```
klauer: To re-cap we have 2 options now:
1. Specify a class
2. Specify a (sub-)module to find classes in

happi/containers.py
```diff
@@ -1,330 +1,27 @@
...
+        registry.update({f"{name}.{var.__name__}": var
+                         for _, var in inspect.getmembers(obj, inspect.isclass)
+                         if issubclass(var, HappiItem)
+                         # Avoid happi internal classes due to imports
+                         and not var.__module__.startswith('happi.')})
```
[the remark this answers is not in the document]
klauer: Also this check of "can we add it to the registry" is a repeated condition that could be made a function: i.e., it's duplicated on line 20 and in this dict comprehension

happi/containers.py
hhslepicka: Correct! I ended up adding support for both.

happi/containers.py
hhslepicka: Will address that.

happi/containers.py
hhslepicka: The second one will be an internal key for Happi at the `registry` and it will point to the proper class `pcdsdevices.happi.containers.Acromag` and queried. I decided to make it shorter as a convenience for users when selecting a class at the CLI but we can also make the namespace shorter at the `pcdsdevices` side and improve the display of the possibilities at the CLI.

klauer: In our example, the name you'd put in a happi entry to get the `Acromag` instantiated would be `pcdsdevices.Acromag`, right?

hhslepicka: That depends but I can see the confusion. Would it be okay if I make the namespace be `pcdsdevices.happi.Acromag` and I switch to use the full class at the registry like you suggested before?

klauer: I'm trying to get at the heart of the problem and don't really know the "right" answer just yet to say yes or no...

`full_module.classname` is unambiguous. Entry_points `name` gives more flexibility, but adds indirection to get to the class itself.

klauer: If we populate the registry with both `entry_point_name.classname` and `module.classname`, I can imagine that'd get confusing as well. I think it has to be one or the other.

The entrypoint name would allow for us to move around classes and refactor codebases without updating our device databases. Maybe that's preferable?

hhslepicka: As of now we have: 
```python
In [1]: import happi
Questionnaire backend unavailable: No module named 'psdm_qs_cli'

In [2]: happi.containers.registry
Out[2]:
{'pcdsdevices.Acromag': pcdsdevices.happi.containers.Acromag,
 'pcdsdevices.AreaDetector': pcdsdevices.happi.containers.AreaDetector,
 'pcdsdevices.Attenuator': pcdsdevices.happi.containers.Attenuator,
 'pcdsdevices.BeamControl': pcdsdevices.happi.containers.BeamControl,
 'pcdsdevices.Diagnostic': pcdsdevices.happi.containers.Diagnostic,
 'pcdsdevices.GateValve': pcdsdevices.happi.containers.GateValve,
 'pcdsdevices.IPM': pcdsdevices.happi.containers.IPM,
 'pcdsdevices.LCLSItem': pcdsdevices.happi.containers.LCLSItem,
 'pcdsdevices.LODCM': pcdsdevices.happi.containers.LODCM,
 'pcdsdevices.Motor': pcdsdevices.happi.containers.Motor,
 'pcdsdevices.MovableStand': pcdsdevices.happi.containers.MovableStand,
 'pcdsdevices.OffsetMirror': pcdsdevices.happi.containers.OffsetMirror,
 'pcdsdevices.PIM': pcdsdevices.happi.containers.PIM,
 'pcdsdevices.PulsePicker': pcdsdevices.happi.containers.PulsePicker,
 'pcdsdevices.Slits': pcdsdevices.happi.containers.Slits,
 'pcdsdevices.Stopper': pcdsdevices.happi.containers.Stopper,
 'pcdsdevices.Trigger': pcdsdevices.happi.containers.Trigger,
 'pcdsdevices.Vacuum': pcdsdevices.happi.containers.Vacuum}
```

So technically we are already doing that, right? Since you have the container class as the value of the dict you can get the `module.classname` ?
I agree, the entrypoint will allow for flexibility on this sense

hhslepicka:
Sorry for the delay... I reworked the registry after talking with Ken.
Also, at the `client` we no longer have the `device_types`. It now points to the registry always and for the `cli.py` we filter-out the deprecated `Device`. It is included at the registry to allow us to properly load old devices that we have.

Most likely the changes here will break the existing `db.json` at `devices-config` repo if the `type` field does not match a key at the registry.


Registry keys are now in the following format:
```
entry_point_name + class_full_qualified_name[1:]
```
This change will allow us to move/rename packages and still be able to find the entries with no changes to the database.

What is not done:
- Aggressive search when a package is passed. E.g.: Look at all python files inside of a package directory.

happi/client.py
```diff
@@ -208,13 +195,11 @@ def create_device(self, device_cls, **kwargs):
...
             device = client.create_device('Device', name='my_device',...)
 
-        See Also
-        --------
-        :attr:`.device_types`
         """
         # If specified by a string
-        if device_cls in self.device_types:
-            device_cls = self.device_types[device_cls]
+        if not inspect.isclass(device_cls) \
```
klauer, line 200: Seems like you could get by without the `inspect.isclass` call

happi/containers.py
```diff
@@ -1,330 +1,122 @@
...
+            module = ".".join(klass.__module__.split(".")[1:])
+            return ".".join([entry_name, module, klass.__name__])
+
+        key = make_entry_name()
+        if key in self._registry:
+            raise RuntimeError(f"Duplicated entry found for key: {key} "
+                               f"and class: {klass}")
+        if klass in self._reverse_registry:
+            dup_key = self._reverse_registry.get(klass)
+            raise RuntimeError(f"Duplicated entry found. Keys: {key} "
```
klauer, line 87: Nice and helpful 👍

klauer:
(Nudging @ZLLentz)

happi/client.py
hhslepicka: That is true... the `device_cls in containers.registry` will take care if it is a string argument.
Addressed via https://github.com/pcdshub/happi/pull/126/commits/d71fabffde7ccadec073e2717023f704cfc82a71

hhslepicka:
@ZLLentz I would gladly nuke that test out but it tests the Questionnaire backend and that directly depends on pcdsdevices to test the type of devices. Sorry about the confusion.

ZLLentz:
@hhslepicka that backend should eventually migrate to a different repo... Surely there's a questionnaire-centered repo somewhere? But it's not an issue so there is no reason to prioritize it

PR #134, opened 2020-04-29
FIX: internally using deprecated all_devices

klauer:
Fixes an annoying deprecation warning.


hhslepicka:
Do you mind to also tackle those two:
https://github.com/pcdshub/happi/blob/205a59415ee7a85ec9768ea315edb9df5d594e7f/happi/client.py#L489

https://github.com/pcdshub/happi/blob/205a59415ee7a85ec9768ea315edb9df5d594e7f/happi/client.py#L721

PR #137, opened 2020-05-19
DOC: Release Notes & other fixes

### LINKED ISSUE #129: Update release notes
Since v1.1.2, we've made releases showing changes here https://github.com/pcdshub/happi/releases but not here https://github.com/pcdshub/happi/blob/master/docs/source/releases.rst.

hhslepicka:
## Description
Docs were broken. This PR fixes the docs to build again and updates the release notes section using https://github.com/pcdshub/pcdsutils/issues/14#issuecomment-630360127 as a tool for it (Thanks, @klauer).

The `autosectionlabel` was a useless extension which was also causing issues with the release notes. I tracked down the instance in which it was being used and replaced it with a regular reference.

## Motivation and Context
Closes #129

## How Has This Been Tested?
Locally

It is documentation.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -47,8 +47,7 @@ extensions = ['sphinx.ext.autodoc',
               'sphinx.ext.napoleon',
               'sphinx.ext.autosummary',
               'IPython.sphinxext.ipython_directive',
-              'IPython.sphinxext.ipython_console_highlighting',
-              'sphinx.ext.autosectionlabel'
+              'IPython.sphinxext.ipython_console_highlighting'
               ]
 
 # In case we don't have the backend libraries installed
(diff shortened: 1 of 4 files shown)


ZryletTC:
What was the issue with `autosectionlabel`? The duplicate label errors for the releases file can be ignored like [this](https://github.com/pcdshub/pcdsdevices/blob/121b7c2935d42972bd617ca3f5fee5ff50206f5e/docs/source/conf.py#L121-L123). I don't see any other errors with it in previous build logs.

hhslepicka:
It was being used in one single place. All others were using pre-existing labels. 
Any reason for keeping it? Does it help with something if we are not using the labels generated?

ZryletTC:
Not really, it's just nice to have sometimes. But if you're not using it, go ahead and remove it.

Issue #142, opened 2020-06-02
CLI happi audit


ZLLentz:
## Expected Behavior

Command to audit the happi database for entries that are not loadable, and help diagnose why. For bonus points, command that helps us migrate these bad entries and make them good.

## Current Behavior

No easy way to debug malformed entries (perhaps from old versions of happi)

## Context

There are devices in the current deployed db at LCLS that are broken and I do not know why

hhslepicka:
Can it be due to the changes we introduced with the moving from happi to pcdsdevices for the containers?

ZLLentz:
This is possible, it's also possible that it's from schema changes that I made.

cristinasewell:
As of now I can insert an OphydItem or HappiItem through the client with extra information, like this: 

```
client = happi.Client(path='../example/db.json')
device_info = {'name': 'alias2',
            'z': 400,
            '_id': 'alias2',
            'prefix': 'BASE:PV',
            'beamline': 'LCLS',
            'type': 'OphydItem',
            'device_class': 'types.SimpleNamespace',
            'args': list(),
            'kwargs': {'hi': 'oh hello'},
            'location_group': 'LOC',
            'functional_group': 'FUNC',
            }
#device = OphydItem(**device_info)
device = HappiItem(**device_info)
client.add_device(device)
```

It adds it fine to the database, and it loads it as well, we can also edit certain fields, but not the ones that the `OphydItem` or `HappiItem` don't have (like `beamline` for ex.) 

This makes it a bit challenging to audit - because it loads the entry fine, yet I think it is a bad entry right?. 
Theoretically, should this type of entry not be allowed to be loaded or do we not want that?

I am wondering if for this audit command we could use something like `jsonschema` and a schema we can check against depending on the item..... I am not sure how complicated it would get though... What are your thoughts on this?

ZLLentz:
I'm not sure if this should be allowed or not. On the one hand, it may be useful to attach extra metadata to your happi entries for specific use cases. On the other hand, this makes it difficult to understand the database from a schemas perspective if any entry can have arbitrary extra keys.

My gut says that the audit shouldn't fail for these right now, and we should continue allowing them, but surely this deserves a warning, something like: "Warning, happi item alias2 has the following extra keys beyond the entry info definition for OphydItem: <keys>. Consider creating a custom happi container to validate these fields."

ZLLentz:
[...]

PR #143, opened 2020-06-02
MNT: don't error out if one of many devices has a bad entry

ZLLentz:
## Description

Wrap the search results in try/except so that one failing device in an `all_devices` call doesn't explode the whole thing.

I'm not sure if this is the best way to fix the problem, but it is a way and it was quick to implement. Maybe this has other negative knock-on effects?

## Motivation and Context

Currently, `typhos <happi-name>` does not work for any device because there is at least one device with a bad entry in lcls's happi config. (There are actually tons of bad entries, but this is irrelevant: https://github.com/pcdshub/device_config/issues/10)

`typhos` has a step where it gets all devices in the database. This fails with a long, confusing traceback about being unable to load the user's device, but this is actually a traceback about failing to load the first device listed in the happi config. This is related to #142.

## How Has This Been Tested?

Ran the tests locally, and they passed, so I didn't break something obvious. I have not yet written a test for this PR.

Not yet

### DIFF
--- a/happi/client.py
+++ b/happi/client.py
@@ -384,8 +384,8 @@ class Client(collections.abc.Mapping):
                 result = wrap_cls(client=self, device=self.find_device(**info))
                 results.append(result)
             except Exception:
-                logger.warning(f'Entry for {info["name"]} is malformed. '
-                               'Skipping.')
+                logger.warning('Entry for %s is malformed. Skipping.',
+                               info['name'])
         return results
 
     def search_range(self, key, start, end=None, **kwargs):
--- a/happi/client.py
+++ b/happi/client.py
@@ -384,7 +384,8 @@ class Client(collections.abc.Mapping):
                 result = wrap_cls(client=self, device=self.find_device(**info))
                 results.append(result)
             except Exception:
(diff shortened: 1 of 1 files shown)


ZLLentz:
I want to make sure Ken has a chance to see this before merging, since he recently did a large refactor in this code base. My test-writing creativity is a bit stumped on this one.

happi/client.py
```diff
@@ -378,8 +378,15 @@ def _get_search_results(self, items, *, wrap_cls=None):
...
         wrap_cls = wrap_cls or self._results_wrap_class
-        return [wrap_cls(client=self, device=self.find_device(**info))
-                for info in items]
+        results = []
+        for info in items:
+            try:
+                result = wrap_cls(client=self, device=self.find_device(**info))
+                results.append(result)
+            except Exception:
+                logger.warning(f'Entry for {info["name"]} is malformed. '
```
klauer, line 387: `f` strings in log messages are generally avoided, as they do their own string interpolation by way of the classic `%s` and can potentially raise exceptions which would be caught by the standard route. Not a big deal though here.

ZLLentz: Right, I forgot about this. Might as well switch it over now.

PR #149, opened 2020-07-06
MNT: handle edge cases in qs loader

ZLLentz:
## Description

The qs database isn't quite as orderly as I would like. There are some edge cases, notably with the xcsdaq13 and xcsdaq18 experiments. This try/except kluge tries to randomwalk us towards the correct entries.

## Motivation and Context

Database is screwy, some of these experiments don't load

## How Has This Been Tested?

Minimally, probably should wait to merge

### DIFF
--- a/happi/backends/qs_db.py
+++ b/happi/backends/qs_db.py
@@ -43,18 +43,24 @@ class QSBackend(JSONBackend):
         # Create our client and gather the raw information from the client
         self.qs = QuestionnaireClient(**kwargs)
 
-        # Ensure that our user entered a valid expname
+        # Get the proposal number
         exp_dict = self.qs.getExpName2URAWIProposalIDs()
         try:
             proposal = exp_dict[expname]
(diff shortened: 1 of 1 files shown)


ZLLentz:
Quick expand on the nature of the edge cases:
```
Info from Urawi for expname -> proposals
'xcslt7717': 'LT77'
'xcsdaq13': 'LDAQ'
<xcsdaq18 missing>

Info from run 18 proposals:
'LV04': {'Instrument': 'CXI', 'proposal_id': 'LV04'}
'xcsdaq13': {'Instrument': 'XCS', 'proposal_id': 'xcsdaq13'}
'xcsdaq18': {'Instrument': 'XCS', 'proposal_id': 'xcsdaq18'}
```

The mapping from URAWI is not guaranteed to be present or correct for these daq experiments
I'm aware my normal experiment examples don't match, but assume there are matching entries for normal experiments

PR #151, opened 2020-07-27
MNT - change db.json examples to use OphydItem 

### LINKED ISSUE #150: example db.json cannot be validated
## Expected Behavior
After loading the example/db.json as a client and running client.validate() nothing should be returned if all devices are properly described.

## Current Behavior
```
In [24]: client = happi.Client(path='./happi/examples/db.json')

In [25]: client.validate()
Failed to validate tv1l0_vgc_1 because The information relating to the device class has been modified to the point where the object can not be initialized, please load the corresponding document
Failed to validate tv1l0_vgc_2 because The information relating to the device class has been modified to the point where the object can not be initialized, please load the corresponding document
Out[25]: ['tv1l0_vgc_1', 'tv1l0_vgc_2']
```

## Possible Solution
Providing a new valid example file for new users of happi, so that they can better understand what is required to build a valid database file

## Environment
Mac OS Catalina 10.15.5
conda environment, 
python 3.8, 
happi 1.6.1, 
ipython 7.16.1,
pymongo 3.10.1
mongomock 3.19.0
matplotlib 3.2.2

cristinasewell:
## Description

- Have removed the previous two devices from `happi/examples/db.json` and added two `OphydItem` items instead (they are using the same names and prefixes as the previous ones)
- Have changed the `happi/happi/qt/model.py` to pass in `entry.item` instead of `entry` when creating an item, and when using the `get_happi_entry_value` function

## Motivation and Context

The example had to be reworked to use OphydItem instead and be standalone.
closes #150 
## How Has This Been Tested?

Check to see if it validates:
```
>>> client = happi.Client(path='./examples/db.json')
>>> client.validate()
[]
```
Run the `listview.py` and `treeview.py` widgets to see the items

This has not been documented

### DIFF
--- a/examples/db.json
+++ b/examples/db.json
@@ -1,52 +1,36 @@
 {
-	"TV1L0-VGC-1": {
-		"_id": "TV1L0-VGC-1",
-		"active": true,
-		"args": [],
-		"beamline": "DEMO_BEAMLINE",
-		"creation": "Tue Feb 27 10:41:25 2018",
-		"device_class": "ophyd.sim.SynAxis",
-		"kwargs": {
-			"name": "{{name}}"
-		},
-		"last_edit": "Thu Apr 12 14:40:08 2018",
-		"macros": null,
-		"name": "tv1l0_vgc_1",
-		"parent": null,
-		"prefix": "TV1L0-VGC-1",
-		"screen": null,
-		"stand": "ST1",
-		"system": "Device",
-		"type": "Vacuum",
-		"z": 1035.5,
-		"invalid": true,
-		"functional_group": "Vacuum",
-		"location_group": "S 1"
-	},
-
-	"TV1L0-VGC-2": {
-		"_id": "TV1L0-VGC-2",
-		"active": true,
-		"args": [],
-		"beamline": "DEMO_BEAMLINE",
-		"creation": "Tue Feb 27 10:41:25 2018",
-		"device_class": "ophyd.sim.SynAxis",
-		"kwargs": {
-			"name": "{{name}}"
-		},
-		"last_edit": "Thu Apr 12 14:40:08 2018",
-		"macros": null,
-		"name": "tv1l0_vgc_2",
-		"parent": null,
-		"prefix": "TV1L0-VGC-2",
-		"screen": null,
-		"stand": "ST1",
-		"system": "Device",
-		"type": "Vacuum",
-		"z": 1035.6,
-		"invalid": true,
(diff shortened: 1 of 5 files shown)


examples/db.json
```diff
@@ -1,52 +1,36 @@
...
+        "type": "OphydItem"
+    },
+    "tv1l0_vgc_2": {
+        "_id": "tv1l0_vgc_2",
+        "active": true,
+        "args": [
+            "{{prefix}}"
+        ],
+        "creation": "Thu Jul 23 17:06:44 2020",
+        "device_class": "ophyd.im.SynAxis",
```
klauer, line 26: Is `validate()` broken? I don't think this should work...
```suggestion
        "device_class": "ophyd.sim.SynAxis",
```

cristinasewell: Thanks for catching that @klauer - I tried it again and it seems to work with `ophyd.im.SynAxis` - maybe I am doing something wrong? Please let me know

```
>>> import happi
>>> client = happi.Client(path='db.json')
>>> client.validate()
[]
>>> client.search(type='OphydItem')
[SearchResult(client=<happi.client.Client object at 0x7f9fbddac850>, metadata={'name': 'tv1l0_vgc_1', 'device_class': 'ophyd.sim.SynAxis', 'args': ['{{prefix}}'], 'kwargs': {'name': '{{name}}'}, 'active': True, 'documentation': None, 'prefix': 'TV1L0-VGC-1', '_id': 'tv1l0_vgc_1', 'creation': 'Thu Jul 23 17:05:48 2020', 'last_edit': 'Thu Jul 23 17:05:48 2020', 'type': 'OphydItem'}), SearchResult(client=<happi.client.Client object at 0x7f9fbddac850>, metadata={'name': 'tv1l0_vgc_2', 'device_class': 'ophyd.im.SynAxis', 'args': ['{{prefix}}'], 'kwargs': {'name': '{{name}}'}, 'active': True, 'documentation': None, 'prefix': 'TV1L0-VGC-2', '_id': 'tv1l0_vgc_2', 'creation': 'Thu Jul 23 17:06:44 2020', 'last_edit': 'Thu Jul 23 17:06:44 2020', 'type': 'OphydItem'})]```

hhslepicka:
@klauer according to `_validate_device` https://github.com/pcdshub/happi/blob/master/happi/client.py#L535 which is called on a loop inside of `validate` it just checks that all fields match the requests from `EntryItem`. I think it is a good idea to expand it to also try to instantiate the device to capture issues like this one.

PR #153, opened 2020-07-28
FIX: fix silent fail when editing a nonexisting field

### LINKED ISSUE #147: BUG: Silent fail when trying to edit a nonexistent field
## Expected Behavior
When attempting to edit a field that an entry doesn't have, the program should warn that it wasn't able to.

## Current Behavior
When doing so, the program continues and acts like everything went properly:
```
(pcds-3.3.0) psbuild-rhel7-01:~ pennebak$ happi edit selector_a beamline=ICL
[2020-06-24 19:25:15] - INFO -  Setting selector_a.beamline = ICL
[2020-06-24 19:25:15] - INFO -  Adding / Modifying information for selector_a ...
+---------------+--------------------------------------+
| EntryInfo     | Value                                |
+---------------+--------------------------------------+
| active        | True                                 |
| args          | ['{{prefix}}']                       |
| creation      | Mon Jun 22 22:54:58 2020             |
| device_class  | pcdsdevices.sample_delivery.Selector |
| documentation | None                                 |
| kwargs        | {'name': '{{name}}'}                 |
| last_edit     | Mon Jun 22 22:54:58 2020             |
| name          | selector_a                           |
| prefix        | ICL:SDS:SEL:A                        |
| type          | OphydItem                            |
+---------------+--------------------------------------+
```

1. Create an entry for an `OphydItem`
2. Try to set the value of a field not present in an `OphydItem` like `beamline`

## Context
Trying to setup happi entries for ICL SDS devices. I now know that I have to use `LCLSItem` instead but wanted to make a note that this should warn when it fails to set a field.

## Your Environment
pcds-3.3.0 on psbuild-rhel7

cristinasewell:
## Description

- I modified `/happi/cli.py` to try to `getattr` when parsing the `edit` argument before setting the object's attribute. If this fails, it logs and error message about the field that was not successfully set. 
- If we try to set any invalid fields, then we exit the process 
 
## Motivation and Context

We want to warn the users when they are trying to edit a non-existing field for a device
We also want to exit the process if any invalid information is detected to keep the integrity of the database 
trying to close #147 

## How Has This Been Tested?

This has been tested with a dummy db.json file 

Test with two fields that fail and one that succeeds: 
`happi edit my_name_here beamline=my_beamline stand=my_stand prefix=my_prefix`
Test with single field that succeeds:
`happi edit my_name_here prefix=my_prefix`

Nowhere 

silent_fail4

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -174,7 +174,7 @@ def happi_cli(args):
                 setattr(device, field, value)
             except Exception:
                 is_invalid_field = True
-                logger.error(f'Could not edit {args.name}.{field}: {ex}')
+                logger.error('Could not edit %s.%s', args.name, field)
         if is_invalid_field:
             sys.exit(1)
         device.save()
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -170,7 +170,7 @@ def happi_cli(args):
             field, value = edit.split('=', 1)
             try:
                 getattr(device, field)
-                logger.info(f'Setting {args.name}.{field} = {value}')
+                logger.info('Setting %s.%s = %s', args.name, field, value)
                 setattr(device, field, value)
             except Exception:
                 is_invalid_field = True
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -165,11 +165,18 @@ def happi_cli(args):
     elif args.cmd == 'edit':
         logger.debug('Starting edit block')
         device = client.find_device(name=args.name)
+        is_edited = False
         for edit in args.edits:
             field, value = edit.split('=', 1)
-            logger.info(f'Setting {args.name}.{field} = {value}')
-            setattr(device, field, value)
-        device.save()
+            try:
+                getattr(device, field)
+                logger.info(f'Setting {args.name}.{field} = {value}')
+                setattr(device, field, value)
+                is_edited = True
+            except AttributeError as ex:
+                logger.warning(f'Could not edit {args.name}.{field}: {ex}')
+        if is_edited:
+            device.save()
         device.show_info()
     elif args.cmd == 'load':
         logger.debug('Starting load block')
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -172,9 +172,9 @@ def happi_cli(args):
                 getattr(device, field)
                 logger.info('Setting %s.%s = %s', args.name, field, value)
                 setattr(device, field, value)
-            except Exception:
+            except Exception as e:
                 is_invalid_field = True
-                logger.error('Could not edit %s.%s', args.name, field)
+                logger.error('Could not edit %s.%s: %s', args.name, field, e)
         if is_invalid_field:
             sys.exit(1)
         device.save()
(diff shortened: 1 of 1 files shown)


cristinasewell:
I am not too confident with this solution.... any ideas or suggestions?

hhslepicka:
Maybe we should fail the whole process if invalid info is present and modify from `warning` to `error`?
Let's also hear from @klauer and @ZLLentz

klauer:
Agreed with @hhslepicka: try to set all attributes and emit an error message. Bailing on the first error is also less than ideal as the rest don't get validated. Also, don't save and then return with an error code (e.g., `sys.exit(1)`)

cristinasewell:
Ok sounds good, thank you for the suggestions. So, just to double check, I should fail the whole process even if some fields are correct - as long as I have one piece of wrong information I should still fail the whole thing?

klauer:
That sounds right, @cristinasewell. Exiting after any detected user error is safest here to keep the integrity of the database, in my opinion.

happi/cli.py
```diff
@@ -165,10 +165,18 @@ def happi_cli(args):
...
+        is_invalid_field = False
         for edit in args.edits:
             field, value = edit.split('=', 1)
-            logger.info(f'Setting {args.name}.{field} = {value}')
-            setattr(device, field, value)
+            try:
+                getattr(device, field)
+                logger.info(f'Setting {args.name}.{field} = {value}')
+                setattr(device, field, value)
+            except AttributeError as ex:
```
klauer, line 175: A bare exception may be better here, as we could catch anything that might go wrong:

```suggestion
            except Exception:
```

happi/cli.py
```diff
@@ -165,10 +165,18 @@ def happi_cli(args):
...
         logger.debug('Starting edit block')
         device = client.find_device(name=args.name)
+        is_invalid_field = False
         for edit in args.edits:
             field, value = edit.split('=', 1)
-            logger.info(f'Setting {args.name}.{field} = {value}')
-            setattr(device, field, value)
+            try:
+                getattr(device, field)
+                logger.info(f'Setting {args.name}.{field} = {value}')
```
klauer, line 173: `f`-strings are great, but they shouldn't be used here.

I know you copied the code that was there, but we should all make a point of fixing this when we see it.
`logger.debug()` et al take a format argument and a list of parameters to format, akin to perhaps `printf` in C.
It performs better (the format string isn't interpolated unless the log message is above the log filtering settings level) and won't crash your program if you messed up the string formatting somehow.

```suggestion
                logger.info('Setting %s.%s = %s', args.name, field, value)
```

happi/cli.py
cristinasewell: Ok sounds good, thank you

happi/cli.py
cristinasewell: Ohh that's good to know, I did not know that. Thank you @klauer

happi/cli.py
klauer: In this case, yes - having the error message is very useful, but catching other exceptions would be good, too. Would you mind changing it to include the error message here?

cristinasewell: Ohh yeah sorry about that!! I added it back.

PR #155, opened 2020-08-12
Cli search regex

### LINKED ISSUE #141: CLI glob/regex search
## Expected Behavior

One of these should help me find the kmono entry:
`happi search kmono`
`happi search *kmono*`
`happi search .*kmono.*`

## Current Behavior

To find the kmono entry, you must precisely request:
`happi search sp1l0_kmono`

## Possible Solution

Use search_regex instead of search
Add a search_glob and use that

### LINKED ISSUE #144: CLI `happi search` does not work with capitalized component names
`happi search AT2L0` does not return anything but `happi search at2l0` works as expected
## Expected Behavior
`happi search AT2L0` should return the object for AT2L0

## Current Behavior
it currently differentiates between capitalized and non-capitalized names. 

## Possible Solution
pass the arg string into a `<string>.lower()`

image

cristinasewell:
## Description
- ` happi search` command line has been modified to use `search_regex` and `search_range` (for z values) instead of the previous `search` function. 

## Motivation and Context
- this change will allow the user to search using the asterisk symbol `*` that matches 0 or more occurrences.
    ex: `*kmono*` will match any occurances that have `kmono` in them
- this also allows the user to search for a range of `z` values by typing a valid range separated by comma like so:
  - z=0,89
**Note:** there is no space between the values

Intended to close #141 
Will possible close #144 as well

## How Has This Been Tested?

Has only been tested locally with items from a dummy json database. 
**Has Not been tested with a mongo database!!** 

**Ex:**
- `happi search z=1,5`
```+--------------------+-------------------------------------+
| EntryInfo          | Value                               |
+--------------------+-------------------------------------+
| active             | True                                |
| args               | ['{{prefix}}']                      |
| beamline           | the_beamline                        |
| creation           | Thu Jul 30 21:48:19 2020            |
| detailed_screen    | None                                |
| device_class       | class                               |
| documentation      | None                                |
| embedded_screen    | None                                |
| engineering_screen | None                                |
| functional_group   | my_func_group                       |
| kwargs             | {'name': '{{name}}'}                |
| last_edit          | Thu Jul 30 23:27:50 2020            |
| lightpath          | False                               |
| location_group     | group                               |
| macros             | None                                |
| name               | the_vacum                           |
| parent             | None                                |
| prefix             | my_prefix                           |
| stand              | None                                |
| system             | vacuum                              |
| type               | pcdsdevices.happi.containers.Vacuum |
| z                  |  4.30000                            |
+--------------------+-------------------------------------+ 
```

- `happi search name=*motor* `

```
+--------------------+------------------------------------+
| EntryInfo          | Value                              |
+--------------------+------------------------------------+
| active             | True                               |
| args               | ['{{prefix}}']                     |
| beamline           | motor_beamline                     |
| creation           | Sun Aug  2 23:07:16 2020           |
| detailed_screen    | motor_screen                       |
| device_class       | motor_clss                         |
| documentation      | docs for motor                     |
| embedded_screen    | motor_embeded_Screen               |
| engineering_screen | motor_engineering_screen           |
| functional_group   | motor_func_group                   |
| kwargs             | {'name': '{{name}}'}               |
| last_edit          | Sun Aug  2 23:07:16 2020           |
| lightpath          | True                               |
| location_group     | motor_group                        |
| macros             | m1                                 |
| name               | motor_name                         |
| parent             | motor_parent                       |
| prefix             | motor_prefi                        |
| stand              | DG1                                |
| system             | motor_system                       |
| type               | pcdsdevices.happi.containers.Motor |
| z                  | 89.00000                           |
+--------------------+------------------------------------+
```
- `happi search name=*name* z=1,90`

```
+--------------------+------------------------------------+
| EntryInfo          | Value                              |
+--------------------+------------------------------------+
| active             | True                               |
| args               | ['{{prefix}}']                     |
| beamline           | motor_beamline                     |
| creation           | Sun Aug  2 23:07:16 2020           |
| detailed_screen    | motor_screen                       |
| device_class       | motor_clss                         |
| documentation      | docs for motor                     |
| embedded_screen    | motor_embeded_Screen               |
| engineering_screen | motor_engineering_screen           |
| functional_group   | motor_func_group                   |
| kwargs             | {'name': '{{name}}'}               |
| last_edit          | Sun Aug  2 23:07:16 2020           |
| lightpath          | True                               |
| location_group     | motor_group                        |
| macros             | m1                                 |
| name               | motor_name                         |
| parent             | motor_parent                       |
| prefix             | motor_prefi                        |
| stand              | DG1                                |
| system             | motor_system                       |
| type               | pcdsdevices.happi.containers.Motor |
| z                  | 89.00000                           |
+--------------------+------------------------------------+
+--------------------+------------------------------------------------------+
| EntryInfo          | Value                                                |
+--------------------+------------------------------------------------------+
| active             | True                                                 |
| args               | ['{{prefix}}']                                       |
| beamline           | pim_beamline                                         |
| creation           | Tue Aug 11 17:03:51 2020                             |
| data               | None                                                 |
| detailed_screen    | None                                                 |
| device_class       | pim_device_class                                     |
| documentation      | None                                                 |
| embedded_screen    | None                                                 |
| engineering_screen | None                                                 |
| functional_group   | pim_func_group                                       |
| kwargs             | {'name': '{{name}}', 'prefix_det': '{{prefix_det}}'} |
| last_edit          | Tue Aug 11 17:03:51 2020                             |
| lightpath          | False                                                |
| location_group     | pim_bamline                                          |
| macros             | None                                                 |
| name               | pim_name                                             |
| parent             | None                                                 |
| prefix             | PIM:IOC                                              |
| prefix_det         | None                                                 |
| stand              | AA9                                                  |
| system             | diagnostic                                           |
| type               | pcdsdevices.happi.containers.PIM                     |
| z                  | 89.34000                                             |
+--------------------+------------------------------------------------------+
```

N/A

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -8,9 +8,6 @@ import fnmatch
 import logging
 import os
 import sys
-import numpy
-import math
-import decimal
 
 import coloredlogs
 from IPython import start_ipython
@@ -92,6 +89,7 @@ def happi_cli(args):
 
         # Get search criteria into dictionary for use by client
         client_args = {}
+        z_is_range = False
         results = []
         for user_arg in args.search_criteria:
             if '=' in user_arg:
@@ -108,35 +106,59 @@ def happi_cli(args):
             if value.replace('.', '').isnumeric():
                 logger.debug('Changed %s to float', value)
                 value = str(float(value))
+
+            # maybe don't check for z - not all of them have z?
+            # but could other fields have a comma? - i think they might
             if criteria == 'z' and ',' in value:
                 start = None
                 end = None
-                val = value.replace('[', '').replace(']', '').split(',')
-                if val[0] != '' and val[0] <= val[1]:
-                    try:
-                        start = decimal.Decimal(val[0])
-                        end = decimal.Decimal(val[1])
-                        dec = max(abs(start.as_tuple().exponent),
-                            abs(end.as_tuple().exponent))
-                        if dec != 0:
-                            step = dec/(10**dec)/dec
-                        else:
-                            dec = 1
-                            step =  dec/(10**dec)/dec
-                        for i in numpy.arange(float(start), float(end) + 1, step):
-                            s = math.floor(i * 10 ** dec) / 10 ** dec
-                            client_args[criteria] = fnmatch.translate(str(s))
-                            results += client.search_regex(**client_args)
-                    except Exception as ex:
-                        logger.info("The value is not numeric or something %s", ex)
-                        sys.exit()
+                z_is_range = True
+                try:
+                    start, end = value.replace(
+                        '[', '').replace(']', '').split(',')
+                    start = float(start)
+                    end = float(end)
+                except Exception as ex:
+                    logger.error("Invalid numbers for the z range %s", ex)
+                    sys.exit()
+                if start < end:
+                    results += client.search_range('z', start, end)
+                else:
+                    logger.error('Invalid z range')
 
-            client_args[criteria] = fnmatch.translate(value)
+            # skip the criteria for range values
+            # it won't be a valid criteria for search_regex()
+            if z_is_range and criteria == 'z':
+                pass
+            else:
+                client_args[criteria] = fnmatch.translate(value)
 
         results += client.search_regex(**client_args)
-        if results:
+
+        # find the repeated items
+        _size = len(results)
+        repeated = []
+        for i in range(_size):
+            k = i + 1
+            for j in range(k, _size):
+                if results[i] == results[j] and results[i] not in repeated:
+                    repeated.append(results[i])
+
+        # we only want to return the ones that have been repeated
+        # they have been matched with both serch_regex() & search_range()
+        if repeated:
+            for res in repeated:
+                res.item.show_info()
+            return repeated
+        # only matched with serach_regex()
+        elif results and not z_is_range:
             for res in results:
-                res.device.show_info()
+                res.item.show_info()
+            return results
+        # only matched with search_range()
+        elif results and z_is_range and len(args.search_criteria) == 1:
+            for res in results:
+                res.item.show_info()
             return results
         else:
             logger.error('No devices found')
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -107,8 +107,6 @@ def happi_cli(args):
                 logger.debug('Changed %s to float', value)
                 value = str(float(value))
 
-            # maybe don't check for z - not all of them have z?
-            # but could other fields have a comma? - i think they might
             if criteria == 'z' and ',' in value:
                 start = None
                 end = None
@@ -120,7 +118,7 @@ def happi_cli(args):
                     end = float(end)
                 except Exception as ex:
                     logger.error("Invalid numbers for the z range %s", ex)
-                    sys.exit()
+                    sys.exit(1)
                 if start < end:
                     results += client.search_range('z', start, end)
                 else:
@@ -136,11 +134,11 @@ def happi_cli(args):
         results += client.search_regex(**client_args)
 
         # find the repeated items
-        _size = len(results)
+        res_size = len(results)
         repeated = []
-        for i in range(_size):
+        for i in range(res_size):
             k = i + 1
-            for j in range(k, _size):
+            for j in range(k, res_size):
                 if results[i] == results[j] and results[i] not in repeated:
                     repeated.append(results[i])
 
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -11,6 +11,7 @@ import sys
 
 import coloredlogs
 from IPython import start_ipython
+from .utils import is_a_range
 
 import happi
 
@@ -109,22 +110,15 @@ def happi_cli(args):
                 logger.debug('Changed %s to float', value)
                 value = str(float(value))
 
-            if (value.replace('.', '').replace(',', '').isnumeric()
-                    and ',' in value):
-                start = None
-                end = None
-                try:
-                    start, end = value.split(',')
-                    start = float(start)
-                    end = float(end)
-                    is_range = True
-                except Exception as ex:
-                    logger.error("Invalid numbers for the range %s", ex)
-                    sys.exit(1)
-                if start < end:
-                    range_list = client.search_range(criteria, start, end)
+            if is_a_range(value):
+                start, stop = value.split(',')
+                start = float(start)
+                stop = float(stop)
+                is_range = True
+                if start < stop:
+                    range_list = client.search_range(criteria, start, stop)
                 else:
-                    logger.error('Invalid range')
+                    logger.error('Invalid range, make sure start < stop')
 
             # skip the criteria for range values
             # it won't be a valid criteria for search_regex()
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -114,11 +114,11 @@ def happi_cli(args):
                     and ',' in value):
                 start = None
                 end = None
-                is_range = True
                 try:
                     start, end = value.split(',')
                     start = float(start)
                     end = float(end)
+                    is_range = True
                 except Exception as ex:
                     logger.error("Invalid numbers for the range %s", ex)
                     sys.exit(1)
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -105,6 +105,9 @@ def happi_cli(args):
                     criteria, value, client_args[criteria]
                 )
                 return
+            if value.replace('.', '').isnumeric():
+                logger.debug('Changed %s to float', value)
+                value = str(float(value))
             if criteria == 'z' and ',' in value:
                 start = None
                 end = None
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -89,9 +89,12 @@ def happi_cli(args):
 
         # Get search criteria into dictionary for use by client
         client_args = {}
-        z_is_range = False
         results = []
+        range_list = []
+        regex_list = []
+        is_range = False
         for user_arg in args.search_criteria:
+            is_range = False
             if '=' in user_arg:
                 criteria, value = user_arg.split('=', 1)
             else:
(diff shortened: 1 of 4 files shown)


cristinasewell:
I would like as much review as possible for this PR :) please - because i used both `search_regex` and `search_range` things got a bit more messy.... maybe i can make it cleaner somehow?
Also I tested this with a dummy json database, but I would like to be able to also test it with a mongo database, do we have to configure the mongo database in a certain way to work with HAPPI - are there any examples or documentation about it?

ZLLentz:
Functionally I like this a lot, I'm finding it helpful and intuitive when navigating the deployed device database. I'll go through the code and make inline comments next.

Regarding mongodb, unfortunately I don't 100% remember how it works or if it needs some sort of special setup. Teddy and I started with a mongodb back at the beginning of happi, but we didn't get the resources to maintain the server instance. So we started using json files since it was easy and because the scale didn't require a database server yet.

happi/cli.py
```diff
@@ -102,13 +105,58 @@ def happi_cli(args):
                 return
             if value.replace('.', '').isnumeric():
                 logger.debug('Changed %s to float', value)
-                value = float(value)
-            client_args[criteria] = value
+                value = str(float(value))
 
-        results = client.search(**client_args)
-        if results:
+            if criteria == 'z' and ',' in value:
```
ZLLentz, line 110: I wonder if there is a way to generalize this more, since in the future we may have other numeric fields that want ranges. Perhaps other position fields? Perhaps other use cases we can't really imagine?

Generalizing this doesn't have to be in-scope for this PR.

happi/cli.py
```diff
@@ -102,13 +105,58 @@ def happi_cli(args):
...
+                start = None
+                end = None
+                z_is_range = True
+                try:
+                    start, end = value.replace(
+                        '[', '').replace(']', '').split(',')
```
ZLLentz, line 116: It's probably ok in cases like this to only accept one form of input- no explicit need to support brackets in addition to no brackets (despite me suggesting it as an option earlier). Most users will soon forget that `z=[0, 250]` is even an option. Since you already went through the trouble though, may as well keep it in.

happi/cli.py
```diff
@@ -102,13 +105,58 @@ def happi_cli(args):
...
+                    results += client.search_range('z', start, end)
+                else:
+                    logger.error('Invalid z range')
+
+            # skip the criteria for range values
+            # it won't be a valid criteria for search_regex()
+            if z_is_range and criteria == 'z':
+                pass
+            else:
+                client_args[criteria] = fnmatch.translate(value)
```
ZLLentz, line 132: I think sticking to the unix glob for the cli was a good call

happi/cli.py
```diff
@@ -102,13 +105,58 @@ def happi_cli(args):
...
+        # find the repeated items
+        res_size = len(results)
+        repeated = []
+        for i in range(res_size):
+            k = i + 1
+            for j in range(k, res_size):
+                if results[i] == results[j] and results[i] not in repeated:
+                    repeated.append(results[i])
+
+        # we only want to return the ones that have been repeated
```
ZLLentz, line 145: This checking for repeats approach assumes that we are doing no more than two searches, so it will need to be refactored if we find ourselves doing more (for example, multiple range searches on different fields)

happi/cli.py
```diff
@@ -102,13 +105,58 @@ def happi_cli(args):
...
+        # they have been matched with both serch_regex() & search_range()
+        if repeated:
+            for res in repeated:
+                res.item.show_info()
+            return repeated
+        # only matched with serach_regex()
```
ZLLentz, line 151: you've got some typos in your comment here and a few lines above

happi/cli.py
ZLLentz: e.g. we'll need to switch to keeping the results from every search separate, then doing the `and` of all the sets

In general, using sets like this would probably clean the combination logic quite a bit

happi/cli.py
cristinasewell: Yeah i agree with you here, i should have a more generalized way to look for ranges. I thought about having only a comma `,` but I don't think it would be sufficient, just in case we have commas in some of our fields.

I need to think about this one....

happi/cli.py
cristinasewell: Yeah I think I need to think about this guy a bit more.... I initially tried to have them in separated lists, but I was kind of running into the same logic mess, because of the multiple scenarios:
1. when using both `search_range` + `search_regex`
    -  I need to return only the ones that they both matched
2. when using only `search_range` 
   - I need to know if i used the `search_regex` as well, if i check for an empty search_regex list is not enough, since i possible did not match anything 
3. when using only `search_regex`
    - here is the same situation as in number 2 - i need an extra variable to know if i have used `search_range` as well...

I also tried to use the sets idea too - something like: `set(a) & set(b)` to give me the ones that are equal in both lists, but I was getting an `unhashable type` error for `SearchResult`  - i though i was just not able to add a dictionary into the set, but I've seen people casting the item into a tuple first, i could give it a try, something like this: 
```
a = {1:2, 3:4}
s = set()
s.add(tuple(a.items()))
```

Maybe I should change my approach here completely... need to think about it - thanks a lot for this feedback 👍

ZLLentz: Interesting, I would have thought that a `SearchResult` would be hashable... That is unfortunate. That sounds like the cleanest way to me, create a list of sets and collapse them with &.

To be clear, I would merge this PR as-is since it is already very useful. Just adding food for thought, anything can be addressed now or we can make a new issue to address it later.

happi/cli.py
ZLLentz: I agree, having a comma is not enough. It's unfortunate that the cli at this point doesn't have a clear API for checking what kind of field a specific criteria is supposed to be, since we may be matching fields across different container classes... If it could just look up what type `criteria` was supposed to be based on the `EntryInfo`, that would be a clean way to decide if we can try to interpret the field as a range.

Something like splitting on the comma and checking for two numeric quantities seems really distasteful but could work...

happi/cli.py
cristinasewell: Sounds good @ZLLentz  - I am trying now to make it a bit cleaner with the use of sets - i agree with you on this, it will look cleaner :)

happi/cli.py
cristinasewell: Yeah i like your idea of being able to look up what type `criteria` was supposed to be based on the `EntryInfo` somehow... for now i basically implemented your suggestion of splitting on the comma and checking - but with a bit of a tweak, please let me know if this is ok: 
instead of checking for two numeric quantities i also split on dot `.` as well (if they are floats) and make a big number out of these two and check if this is numeric..... am I covering all the scenarios here, what do you think? :)

happi/cli.py
cristinasewell: I took this option out, I think you are right that soon people will forget about `[0,250]` being an option - also because i took it out made things a bit cleaner.... please let me know if I should add it back in

happi/cli.py
cristinasewell: Yeah thanks for this idea, i think you mentioned it in the Issue description and @klauer also elaborated on it - thank you, i like this too

happi/cli.py
cristinasewell: thanks for catching that

happi/cli.py
cristinasewell: Still working on this one ... i am sorry it takes me so long to get this figured out, It might be that I don't know what i am doing here, I can't get those items in a `Set` like I wanted (casting into tuples) - and I am trying to come up with a cleaner way i can do that with but anything i try gets messy, my brain is not thinking well - please give me a bit more time on this and hopefully i'll come up with something a bit cleaner ?

cristinasewell:
@ZLLentz  I tried to cleanup a bit the 'search' portion from `cli.py` by moving a couple little general functions in `utils.py` - please let me know if that's fine or if I made it worse.

happi/cli.py
cristinasewell: I changed this a bit in my last commit 8caebba - please let me know what you think

happi/utils.py
```diff
@@ -20,3 +23,31 @@ def get_happi_entry_value(entry, key, search_extraneous=True):
...
+        1,100
+        -2,8
+        100,2
+    """
+    if ',' in str_value:
+        start, stop = str_value.split(',')
+        if is_number(start) and is_number(stop):
+            return True
+        else:
+            logger.error("Possibly provided invalid numbers for a range")
```
ZLLentz, line 53: We probably want to return `False` at the bottom here instead of `None`

cristinasewell: Yes that's right, thank you for catching that. Fixed it in cf7fd10

ZLLentz: Still one more hole here, where were return `None` if the initial `if ',' in str_value` is `False`

cristinasewell: Ohh sorry about that, added it in bf1c57f .

ZLLentz:
I will merge now unless you think I should wait

cristinasewell:
@ZLLentz  I am usually a paranoid person when it comes to my code braking something important hahaha ... so I will tend to say wait....   but in this case I think it would be useful if more people would take a spin at it and try it out and give us some feedback (either before or after merging it - depending on how busy people are, it seems like people are really busy all the time)
I think it works fine, but because I still don't have a lot of experience with HAPPI as others do, I could be missing things.... 

The fact that I had to use two different search functions (`search_range` and `search_regex`) made the implementation a bit more messy, but I don't think there is much to do there now unless we change some major things. 
Also, maybe for the future we can have the commands (search, add, ...) in separated files/classes to make it cleaner... 

It is up to you

ZLLentz:
I'm confident this is an improvement, so I'm going to merge now to get it into the next release. If there are problems that pop up later it is ok and we can iterate. This only touches the cli and not the core (aside from the str cast) so the risk is extremely low.

Issue #96, opened 2019-05-16
Happi Database restructuring


teddyrendahl:
# Problem
Had a revelation about how to restructure the `happi` database. Right now when the database is written we structure the JSON dump as a single large dictionary, the key is determined by `Client._id`. This is the only key that `happi` checks to be unique

# Restructuring
Add a keyword to a piece of `EntryInfo` called `unique`. This means that this piece of information has to be unique among all entries (i.e PV prefix, name, e.t.c). The database is then stored as a flat list of dictionaries, but when when we load it in `happi.Client` we can create a look-up table for all of our unique keys. This will give us fast look-up on any unique key and we can remove the `_id` nonsense.

PR #157, opened 2020-08-17
MNT: fix entry renaming

### LINKED ISSUE #131: Cannot change device name field using json backend
## Expected Behavior
Instantiating a `happi.Client` in an IPython terminal, one may try to edit a device name as follows:
```
import happi
client = happi.Client(path='/reg/g/pcds/pyps/apps/hutch-python/device_config/db.json')
dev = client.find_device(name='mr1l0')
dev.name = 'mr1l0_homs'
dev.save()
```
## Current Behavior
Saving the device as shown above (last line) results in the following error:
```
In [14]: dev.save()
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/reg/g/pcds/pyps/apps/dev/pythonpath/happi/backends/json_db.py in save(self, _id, post, insert)
    280                 try:
--> 281                     db[_id].update(post)
    282                 except KeyError:

KeyError: 'mr1l0_homs'

During handling of the above exception, another exception occurred:

SearchError                               Traceback (most recent call last)
<ipython-input-14-f5fa237e2afe> in <module>
----> 1 dev.save()

/reg/g/pcds/pyps/apps/dev/pythonpath/happi/client.py in <lambda>()
    290
    291         # Add the save method to the device
--> 292         device.save = lambda: self._store(device, insert=False)
    293         return device
    294

/reg/g/pcds/pyps/apps/dev/pythonpath/happi/client.py in _store(self, device, insert)
    594         # Store information
    595         logger.info('Adding / Modifying information for %s ...', _id)
--> 596         self.backend.save(_id, post, insert=insert)
    597         return _id
    598

/reg/g/pcds/pyps/apps/dev/pythonpath/happi/backends/json_db.py in save(self, _id, post, insert)
    281                     db[_id].update(post)
    282                 except KeyError:
--> 283                     raise SearchError("No device found {}".format(_id))
    284
    285     def delete(self, _id):

SearchError: No device found mr1l0_homs
```
This works for all other happi fields, but not for `name` likely due to a connection with the `_id` field.
Instead, to change the device name one must simply delete the happi entry and re-add it back in with the new name.

## Possible Solution
It seems like the `SearchError` generated by the `save()` function call is due to searching the `new_name` before the happi entry is updated. It may be as simple as having a variable `old_name` that can be used for searching the database.

## Your Environment
Using `dev_conda` environment:
`/reg/g/pcds/pyps/conda/dev_conda`

cristinasewell:
## Description

`client.py`'s **`_store`** method has been modified to store the old_name to be able to find the entry in case the user is trying to change its name. An extra check has been added to see if the user is trying to change the name of an existing entry, if so, a new entry is created with a new name and the older entry gets deleted after that. 

## Motivation and Context

These changes allow the users to change the 'name' of an entry. 
Because the 'name' is tied to the '_id' of an entry (as of now, this might change in the future), it is not practicable to change only the name for an entry, the entire entry has to be removed and replaced with the new one containing a new name. 

This is trying to close #131 

## How Has This Been Tested?

Has only been tested locally and with a dummy json database. 

Contents of json database before running the test - trying to update the name of **`an_entry`**:
```
{
    "a_name": {
        "_id": "a_name",
        "active": true,
        "args": [
            "{{prefix}}"
        ],
        "creation": "Thu Aug 13 00:28:59 2020",
        "device_class": "class",
        "documentation": null,
        "kwargs": {
            "name": "{{name}}"
        },
        "last_edit": "Mon Aug 17 14:06:49 2020",
        "name": "a_name",
        "prefix": "PREFIX",
        "type": "OphydItem"
    },
    "an_entry": {
        "_id": "an_entry",
        "active": true,
        "args": [
            "{{prefix}}"
        ],
        "creation": "Thu Aug 13 11:01:48 2020",
        "device_class": "the_device_class",
        "documentation": null,
        "kwargs": {
            "name": "{{name}}"
        },
        "last_edit": "Mon Aug 17 14:18:00 2020",
        "name": "an_entry",
        "prefix": "THE:PREFIX",
        "type": "OphydItem"
    },
...
```
```
import happi

client = happi.Client(path='../example/db.json')
item = client.find_device(name='an_entry')
item.name = 'an_entry_test'

item.show_info()
item.save()
```
Contents of json database after test ran:

```
{
    "a_name": {
        "_id": "a_name",
        "active": true,
        "args": [
            "{{prefix}}"
        ],
        "creation": "Thu Aug 13 00:28:59 2020",
        "device_class": "class",
        "documentation": null,
        "kwargs": {
            "name": "{{name}}"
        },
        "last_edit": "Mon Aug 17 14:06:49 2020",
        "name": "a_name",
        "prefix": "PREFIX",
        "type": "OphydItem"
    },
    "an_entry_test": {
        "_id": "an_entry_test",
        "active": true,
        "args": [
            "{{prefix}}"
        ],
        "creation": "Thu Aug 13 11:01:48 2020",
        "device_class": "the_device_class",
        "documentation": null,
        "kwargs": {
            "name": "{{name}}"
        },
        "last_edit": "Mon Aug 17 14:39:35 2020",
        "name": "an_entry_test",
        "prefix": "THE:PREFIX",
        "type": "OphydItem"
    },
    "an_item": {
        "_id": "an_item",
        "active": true,
        "args": [
....
```

N/A

### DIFF
--- a/happi/client.py
+++ b/happi/client.py
@@ -607,7 +607,6 @@ class Client(collections.abc.Mapping):
         try:
             if the_old_name:
                 _id = the_old_name
-                post['_id'] = _id
             else:
                 _id = post[self._id_key]
         except KeyError:
--- a/happi/client.py
+++ b/happi/client.py
@@ -608,19 +608,22 @@ class Client(collections.abc.Mapping):
         try:
             # In case we want to update the name of an entry
             # We want to add a new entry, and delete the old one
+            _id = post[self._id_key]
             if the_old_name != post[self._id_key]:
-                _id = the_old_name
-                self.backend.save(post[self._id_key], post, insert=True)
+                # Store information for the new entry
+                logger.info('Saving new entry %s ...', _id)
+                self.backend.save(_id, post, insert=True)
+                # Remove the information for the old entry
+                logger.info('Removing old entry %s ...', the_old_name)
                 self.backend.delete(the_old_name)
             else:
-                _id = post[self._id_key]
+                # Store information
+                logger.info('Adding / Modifying information for %s ...', _id)
                 self.backend.save(_id, post, insert=insert)
         except KeyError:
             raise EntryError('HappiItem did not supply the proper information '
                              'to interface with the database, missing {}'
                              ''.format(self._id_key))
-        # Store information
-        logger.info('Adding / Modifying information for %s ...', _id)
         return _id
 
     @classmethod
--- a/happi/client.py
+++ b/happi/client.py
@@ -609,7 +609,7 @@ class Client(collections.abc.Mapping):
             # In case we want to update the name of an entry
             # We want to add a new entry, and delete the old one
             _id = post[self._id_key]
-            if the_old_name != post[self._id_key]:
+            if the_old_name and the_old_name != post[self._id_key]:
                 # Store information for the new entry
                 logger.info('Saving new entry %s ...', _id)
                 self.backend.save(_id, post, insert=True)
--- a/happi/client.py
+++ b/happi/client.py
@@ -606,24 +606,24 @@ class Client(collections.abc.Mapping):
 
         # Find id
         try:
-            # In case we want to update the name of an entry
-            # We want to add a new entry, and delete the old one
             _id = post[self._id_key]
-            if the_old_name and the_old_name != post[self._id_key]:
-                # Store information for the new entry
-                logger.info('Saving new entry %s ...', _id)
-                self.backend.save(_id, post, insert=True)
-                # Remove the information for the old entry
-                logger.info('Removing old entry %s ...', the_old_name)
-                self.backend.delete(the_old_name)
-            else:
-                # Store information
-                logger.info('Adding / Modifying information for %s ...', _id)
-                self.backend.save(_id, post, insert=insert)
         except KeyError:
             raise EntryError('HappiItem did not supply the proper information '
                              'to interface with the database, missing {}'
                              ''.format(self._id_key))
+        # In case we want to update the name of an entry
+        # We want to add a new entry, and delete the old one
+        if the_old_name and the_old_name != post[self._id_key]:
+            # Store information for the new entry
+            logger.info('Saving new entry %s ...', _id)
+            self.backend.save(_id, post, insert=True)
+            # Remove the information for the old entry
+            logger.info('Removing old entry %s ...', the_old_name)
+            self.backend.delete(the_old_name)
+        else:
+            # Store information
+            logger.info('Adding / Modifying information for %s ...', _id)
+            self.backend.save(_id, post, insert=insert)
         return _id
 
     @classmethod
--- a/happi/client.py
+++ b/happi/client.py
@@ -584,6 +584,9 @@ class Client(collections.abc.Mapping):
         self._validate_device(device)
         # Grab information from device
         post = device.post()
+        # save the old name in case the user is trying to
+        # change the 'name' of an entry
+        the_old_name = post.get('_id', None)
         # Store creation time
         creation = post.get('creation', ttime.ctime())
         # Clean supplied information
@@ -602,7 +605,11 @@ class Client(collections.abc.Mapping):
                      'last_edit': ttime.ctime()})
         # Find id
         try:
-            _id = post[self._id_key]
+            if the_old_name:
+                _id = the_old_name
+                post['_id'] = _id
+            else:
+                _id = post[self._id_key]
         except KeyError:
             raise EntryError('HappiItem did not supply the proper information '
                              'to interface with the database, missing {}'
--- a/happi/client.py
+++ b/happi/client.py
@@ -603,19 +603,24 @@ class Client(collections.abc.Mapping):
         post.update({'type': tpe,
                      'creation': creation,
                      'last_edit': ttime.ctime()})
+
         # Find id
         try:
-            if the_old_name:
+            # In case we want to update the name of an entry
+            # We want to add a new entry, and delete the old one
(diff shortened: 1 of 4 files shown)


cristinasewell:
I am not too keen on this approach - but i can't think of a better one now. #96 would probably solve this problem better.... 
Any feedback or suggestions here please? 

Also, I only implemented this for the json database...

ZLLentz:
I think the approach is fine, a full rework is a bit much to do right now. I'm a bit nervous that we needed to make changes to the json backend, but maybe only the json backend was affected? I need to add an issue for making setup instructions for the mongodb.

cristinasewell:
Ohhh maybe I could have done everything on the client side..... i will give that a try

cristinasewell:
@ZLLentz  i changed the implementation so now everything is handled in `client.py` - please let me know what you think

happi/client.py
```diff
@@ -584,6 +584,9 @@ def _store(self, device, insert=False):
         self._validate_device(device)
         # Grab information from device
         post = device.post()
+        # save the old name in case the user is trying to
+        # change the 'name' of an entry
+        the_old_name = post.get('_id', None)
```
klauer, line 589:
```suggestion
        the_old_name = post.get(self._id_key, None)
```

happi/client.py
```diff
@@ -600,16 +603,24 @@ def _store(self, device, insert=False):
...
                      'creation': creation,
                      'last_edit': ttime.ctime()})
+
         # Find id
         try:
-            _id = post[self._id_key]
+            # In case we want to update the name of an entry
+            # We want to add a new entry, and delete the old one
+            if the_old_name != post[self._id_key]:
+                _id = the_old_name
```
klauer, line 612: The returned `_id` should be the valid one, not the returned one.

happi/client.py
```diff
@@ -600,16 +603,24 @@ def _store(self, device, insert=False):
...
+                self.backend.delete(the_old_name)
+            else:
+                _id = post[self._id_key]
+                self.backend.save(_id, post, insert=insert)
         except KeyError:
             raise EntryError('HappiItem did not supply the proper information '
                              'to interface with the database, missing {}'
                              ''.format(self._id_key))
         # Store information
         logger.info('Adding / Modifying information for %s ...', _id)
```
klauer, line 623: The logger message should be moved up to the individual `save` and `delete` calls, indicating what's happening. e.g.,
`Saving new entry...`
`Removing old entry...`

klauer:
Handling it with the backend API and not in the backend itself was the right call. 👍

happi/client.py
cristinasewell: Thanks @klauer  for the suggestion. 
The `post.get(self_id_key)` is already the `new_name` here...  so I don't think I can rely on this one, unless I make other changes maybe, or unless i can get it from somewhere else?

klauer: Sorry, this suggestion shouldn't have been added. Please ignore it.

klauer:
It's unclear why, but GitHub dropped one of my comments explaining just how confusing the current setup is, but why @cristinasewell's code appears correct and is working.

I think the gist of my previous message was along the lines of:
* `self._id_key` is not `_id` as you might expect, it's `name`.
* The user requests changes to the `name` key.
* The "old name" is grabbed from `_id`, and the updated name is in `post['name']`
* So, `post['name'] != post['_id']`, causing the save/delete clause to be evaluated.
* Then, `post['_id']` is updated as a client key to `post['name']` (this happens in the backend itself)

happi/client.py
cristinasewell: Ok that makes sense, thank you - `the_old_name` won't be valid after this anymore... 
Fixed it in b69e45d - basically i am just returning `post[self._id_key]` ?? please let me know if that's not what you meant

happi/client.py
cristinasewell: Fixed this in b69e45d

happi/client.py
klauer: Another typo in my comment, apologies for the distracted commenting. I meant "not the ~returned~ deleted one"

That's correct 👍

cristinasewell:
Just realized that I think i broke the `add` method - converting this to draft until I fix it

cristinasewell:
@klauer regarding your explanation above for the current confusing setup (thanks for that btw)  - do you think we could/should  handle the "changing of the name process" differently somehow? 

Also i pushed a fix in 84e0142 - where I am checking if I have an `old_name` in the first place

happi/client.py
```diff
@@ -600,16 +603,27 @@ def _store(self, device, insert=False):
         post.update({'type': tpe,
                      'creation': creation,
                      'last_edit': ttime.ctime()})
+
         # Find id
         try:
+            # In case we want to update the name of an entry
+            # We want to add a new entry, and delete the old one
             _id = post[self._id_key]
+            if the_old_name and the_old_name != post[self._id_key]:
```
klauer, line 612: I think you should move this block out of the `try/except KeyError`, as we're only trying to see if `post[id_key]` exists.

cristinasewell: Sounds good, moved it in 55a83ad

cristinasewell:
> I think the renaming isn't bad, it's just always been strange that `_id_key == 'name'` when there's a key called `_id`

I agree with that, it is strange - i am not sure why we have it this way

ZLLentz:
:shrug: This code has undergone many revisions, so it could be a relic of past versions.

I'm going to merge this fix and include it in the next release.

ZLLentz:
I reread the code one last time- this final version is very clean. I like it.

PR #159, opened 2020-08-18
DOC: add release notes for v1.7.0

ZLLentz:
Includes all the recent updates/fixes that @cristinasewell worked on.
We can also include #157 here, but it is ok if that waits for the next tag.

### DIFF
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -2,6 +2,24 @@
  Release History
 =================
 
+v1.7.0 (2020-08-18)
+===================
+
+Features
+--------
+- Add cli search globbing, e.g. now the following will work:
+  ``happi search xpp*`` (show all devices whose names start with xpp)
+
+Bugfixes
+--------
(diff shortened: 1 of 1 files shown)


docs/source/releases.rst
```diff
@@ -2,6 +2,24 @@
...
+===================
+
+Features
+--------
+- Add cli search globbing, e.g. now the following will work:
+  ``happi search xpp*`` (show all devices whose names start with xpp)
+
+Bugfixes
+--------
+- Fix issue with silent failure when editing a non-existant field.
```
klauer, line 15:
```suggestion
- Fix issue with silent failure when editing a non-existent field.
```

ZLLentz: slip of the keyboard, you know

ZLLentz:
Thank you for all your efforts here @cristinasewell, I can never find time to work on happi so I'm glad to see these come through

PR #160, opened 2020-08-19
ENH: Output usage without subcommands

### LINKED ISSUE #154: `happi` without any sub-commands outputs nothing
## Current Behavior

Typing `happi` with no sub-commands outputs nothing:
```bash
$ happi
$
```

## Expected Behavior
It should output, say, the help information:

```
$ happi
usage: happi [-h] [--path PATH] [--verbose] [--version]
             {search,add,edit,load} ...
```

cristinasewell:
## Description

Using `len(sys.argv)` to count the arguments passed to happi, and printing the happi usage if only one argument is passed

## Motivation and Context

This change allows the user to see what commands are available when only `happi` is typed without any sub commands. 

Trying to close #154 

## How Has This Been Tested?

This has been tested locally only: 
```
(python3.7) $ happi
usage: happi [-h] [--path PATH] [--verbose] [--version]
             {search,add,edit,load} ...
```

N/A

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -64,7 +64,7 @@ def happi_cli(args):
     parser = get_parser()
     args = parser.parse_args(args)
 
-    # print how to use happi if no argument provided
+    # print happi usage if no arguments are provided
     if len(sys.argv) == 1:
         parser.print_usage()
 
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -62,12 +62,11 @@ def get_parser():
 
 def happi_cli(args):
     parser = get_parser()
-    args = parser.parse_args(args)
-
     # print happi usage if no arguments are provided
-    if len(sys.argv) == 1:
+    if not args:
         parser.print_usage()
         return
+    args = parser.parse_args(args)
 
     # Logging Level handling
     if args.verbose:
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -67,6 +67,7 @@ def happi_cli(args):
     # print happi usage if no arguments are provided
     if len(sys.argv) == 1:
         parser.print_usage()
+        sys.exit(0)
 
     # Logging Level handling
     if args.verbose:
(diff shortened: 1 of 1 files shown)


happi/cli.py
```diff
@@ -64,6 +64,10 @@ def happi_cli(args):
     parser = get_parser()
     args = parser.parse_args(args)
 
+    # print happi usage if no arguments are provided
+    if len(sys.argv) == 1:
+        parser.print_usage()
+
```
klauer, line 70: I think it should probably return here after printing the usage.

cristinasewell: I added `sys.exit(0)` in 5bbfc78 - Is this what you meant @klauer ? or should i use a different exit code, or an actual return?

klauer: The problem with `sys.exit` is that it's a heavy-handed solution and doesn't respect the fact that the caller may not want the process to exit. Returning is, in my opinion, most appropriate here.

cristinasewell: Ok thank you, that sounds good, changed it in 8826d6d

cristinasewell: i'm failing the build though, let me take a look

klauer: Strange - try `pytest -vv --pdb` with/without the `-s` and look at the backtrace.

cristinasewell: The problem for me is that as soon as I add some parameters like `-vv` or `--pdb` I don't get the errors anymore, so can't look at the backtrack - hahah - I don't get it. 
I change the implementation in 4d6cc28 and it works this way... but I still want to know why it was not working the other way.

klauer: Oh, looks like because `sys.argv` is getting your `pytest` arguments and not `happi` CLI arguments. That'll do it for sure.

klauer:
I see this is still a draft, but it looks ready to me. Merge if ready.

PR #161, opened 2020-08-19
FIX: fix typo in cli.py for clone command

### LINKED ISSUE #156: Cli --clone argument does not work
## Expected Behavior

We expect this argument to help us create a new entry that is very similar to an old entry, but change only a few fields.

## Current Behavior

https://github.com/pcdshub/happi/issues/131#issuecomment-673669346

1. `happi --clone <devicename>`

cristinasewell:
## Description

It turns out there was just a simple typo when calling a function from `HappiRegistry`. It might have been that the function itself was changed after the clone was added, maybe not... 
So this PR fixed this typo and now the `clone` command seems to work fine. 
 
## Motivation and Context

This change was necessary to make the `clone` command work. 

Trying to close #156 

## How Has This Been Tested?

Has been tested locally with a dummy json database: 
Trying cloning **`an_item`**:
```
happi add --clone an_item

[2020-08-19 09:45:36] - INFO -  Enter value for name, default=an_item, enforce=re.compile('[a-z][a-z\\_0-9]{2,78}$')
an_item_2
[2020-08-19 09:45:45] - INFO -  Enter value for device_class, default=class, enforce=<class 'str'>

[2020-08-19 09:45:48] - INFO -  Selecting default value class
[2020-08-19 09:45:48] - INFO -  Enter value for args, default=['{{prefix}}'], enforce=<class 'list'>

[2020-08-19 09:45:49] - INFO -  Selecting default value ['{{prefix}}']
[2020-08-19 09:45:49] - INFO -  Enter value for kwargs, default={'name': '{{name}}'}, enforce=<class 'dict'>

[2020-08-19 09:45:50] - INFO -  Selecting default value {'name': '{{name}}'}
[2020-08-19 09:45:50] - INFO -  Enter value for active, default=True, enforce=<class 'bool'>

[2020-08-19 09:45:51] - INFO -  Selecting default value True
[2020-08-19 09:45:51] - INFO -  Enter value for documentation, default=None, enforce=<class 'str'>

[2020-08-19 09:45:52] - INFO -  Selecting default value None
[2020-08-19 09:45:52] - INFO -  Enter value for prefix, default=PREFIX, enforce=<class 'str'>

[2020-08-19 09:45:53] - INFO -  Selecting default value PREFIX
[2020-08-19 09:45:53] - INFO -  Please confirm the following info is correct:
+---------------+----------------------+
| EntryInfo     | Value                |
+---------------+----------------------+
| active        | True                 |
| args          | ['{{prefix}}']       |
| device_class  | class                |
| documentation | None                 |
| kwargs        | {'name': '{{name}}'} |
| name          | an_item_2            |
| prefix        | PREFIX               |
+---------------+----------------------+
y/N
y
[2020-08-19 09:45:58] - INFO -  Adding device
[2020-08-19 09:45:58] - INFO -  Storing device OphydItem (name=an_item_2) ...
[2020-08-19 09:45:58] - INFO -  Adding / Modifying information for an_item_2 ...
[2020-08-19 09:45:58] - INFO -  HappiItem OphydItem (name=an_item_2) has been succesfully added to the database
```

Contents of database after this change: 
```
    "an_item": {
        "_id": "an_item",
        "active": true,
        "args": [
            "{{prefix}}"
        ],
        "creation": "Thu Aug 13 00:28:59 2020",
        "device_class": "class",
        "documentation": null,
        "kwargs": {
            "name": "{{name}}"
        },
        "last_edit": "Mon Aug 17 15:45:15 2020",
        "name": "an_item",
        "prefix": "PREFIX",
        "type": "OphydItem"
    },
    "an_item_2": {
        "_id": "an_item_2",
        "active": true,
        "args": [
            "{{prefix}}"
        ],
        "creation": "Wed Aug 19 09:45:58 2020",
        "device_class": "class",
        "documentation": null,
        "kwargs": {
            "name": "{{name}}"
        },
        "last_edit": "Wed Aug 19 09:45:58 2020",
        "name": "an_item_2",
        "prefix": "PREFIX",
        "type": "OphydItem"
    },
```

N/A

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -163,7 +163,7 @@ def happi_cli(args):
         if args.clone:
             clone_source = client.find_device(name=args.clone)
             # Must use the same container if cloning
-            response = registry.entry_from_class(clone_source.__class__)
+            response = registry.entry_for_class(clone_source.__class__)
         else:
             # Keep Device at registry for backwards compatibility but filter
             # it out of new devices options


cristinasewell:
Yeah i was thinking about adding a test and then i changed my mind because I saw that we don't have a test for the `add` command either - not a good excuse haha
Do you have anything to recommend for argparse with user prompts?
I saw an example here: https://stackoverflow.com/questions/48359957/pytest-with-argparse-how-to-test-user-is-prompted-for-confirmation that looked similar to what I need, what do you think? or am I complicating this?

klauer:
Ah, I hadn't realized there was no `add` one either, but I can see why. 

It isn't necessary to add this (and is really our own fault that it isn't there already), but if you're interested in giving it a try it would be a good thing for us to have some coverage here.

It would require you to do as the stackoverflow post indicates - patching the builtin `input` with your own method.
And, unfortunately, we don't use `input()` with text that you could look for. That'd mean you'd have to use `caplog` to look for the latest log message to determine what the test should reply with.

So if you record log messages, you'd see something like this for the first `add`: `[2020-08-19 12:02:46] - INFO -  Please select a container, or press enter for generic Ophyd Device container:`
Then `input()` would be called with no arguments, so you'd patch it to return the container type string. And repeat for the remaining prompts.

We should make an issue out of the lack of coverage/tests for the CLI, regardless.

cristinasewell:
Yeah I agree, I'd like to try adding the tests - that would be a good practice for me too. 
Should I create a new Issue or should I try adding it here? - it might just take a bit longer until I figure it out and make it work, but if we are not in a rush I can add the tests here.

klauer:
Either way is fine - I have no problems merging this as-is. 👍

cristinasewell:
@klauer I submitted an issue where we can address the tests for all the other commands as well. #162  - so we can merge this one then - this way we'll keep the problems separately 
I will work on #162

PR #164, opened 2020-08-22
TST: Cli testing

### LINKED ISSUE #162: Lack of some coverage/tests for CLI
## Expected Behavior

Some testing for the `add`, `clone`, `edit`, `load` commands would be great.

## Current Behavior

Currently those commands are not tested 
## Possible Solution

https://github.com/pcdshub/happi/pull/161#issuecomment-676606588
## Context

 This would allow programmers to refactor code and make sure the module still works correctly.

cristinasewell:
## Description

Added more test cases for some cli functionalities 
## Motivation and Context

Trying to close #162 

## How Has This Been Tested?

```
Has only been tested locally: 

python run_tests.py
coverage run run_tests.py 
coverage report -m
```

N/A

### DIFF
--- a/happi/tests/test_cli.py
+++ b/happi/tests/test_cli.py
@@ -1,8 +1,11 @@
 # test_cli.py
 
+import builtins
+import logging
 import pytest
 import happi
 from happi.cli import happi_cli
+from unittest import mock
 
 
 @pytest.fixture(scope='function')
@@ -79,12 +82,37 @@ def test_cli_version(capsys):
     assert happi.__file__ in readout.out
 
 
+def test_cli_no_argument(capsys):
+    happi.cli.happi_cli([])
+    readout = capsys.readouterr()
+    # TODO: figure out a way to get usage: happi
+    # instead of usage: pytest or usage run_test.py..
+    assert 'usage:' in readout.out
+    assert '[-h] [--path PATH] [--verbose] [--version]\n'
+    '                    {search,add,edit,load} ...\n' in readout.out
+
+
 def test_search(happi_cfg):
     client = happi.client.Client.from_config(cfg=happi_cfg)
     res = client.search_regex(beamline="TST")
     res_cli = happi.cli.happi_cli(['--verbose', '--path', happi_cfg, 'search',
                                    'beamline=TST'])
     assert [r.device for r in res] == [r.device for r in res_cli]
+
+
+def test_search_with_name(happi_cfg, caplog):
+    client = happi.client.Client.from_config(cfg=happi_cfg)
+    res = client.search_regex(name='TST_BASE_PIM2')
+    res_cli = happi.cli.happi_cli(['--verbose', '--path', happi_cfg, 'search',
+                                   'TST_BASE_PIM2'])
+    assert [r.device for r in res] == [r.device for r in res_cli]
+    # test duplicate search parameters
+    with caplog.at_level(logging.ERROR):
+        res_cli = happi.cli.happi_cli(['--verbose', '--path', happi_cfg,
+                                       'search', 'TST_BASE_PIM2',
+                                       'name=TST_BASE_PIM2'])
+        assert "Received duplicate search criteria" in caplog.text
+        caplog.clear()
 
 
 def test_search_z(happi_cfg):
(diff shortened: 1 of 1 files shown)


happi/tests/test_cli.py
```diff
@@ -95,9 +124,209 @@ def test_search_z(happi_cfg):
...
+def test_load(caplog, happi_cfg, capsys):
+    device_info = ['HappiItem', 'happi_name', 'types.SimpleNamespace',
+                   [], {'name': 'my_name'}, True, 'docs', 'y']
+    with mock.patch.object(
+            builtins, 'input', lambda x=None: device_info.pop(0)):
+        # add device first
+        happi.cli.happi_cli(['--verbose', '--path', happi_cfg, 'add'])
+    caplog.clear()
+    # try to load the device
+    # TODO - try somewhow to mock the start_ipython or close the shell
```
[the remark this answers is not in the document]
cristinasewell: I am not too happy with how I tested the `load` command ... as I am not testing the `load_devices` function (although we are testing this in `test_client.py`)  - but I rather tried to test if the `start_ipython` was called with the proper attributes - should I be mocking the `load_devices` function here as well - any advice what is the best approach here?

happi/tests/test_cli.py
```diff
@@ -79,6 +83,16 @@ def test_cli_version(capsys):
...
 
 
+def test_cli_no_argument(capsys):
+    happi.cli.happi_cli([])
+    readout = capsys.readouterr()
+    # TODO: figure out a way to get usage: happi
+    # instead of usage: pytest or usage run_test.py..
+    assert 'usage:' in readout.out
+    assert '[-h] [--path PATH] [--verbose] [--version]\n'
+    '                    {search,add,edit,load} ...\n' in readout.out
```
klauer, line 93: I think this isn't part of the assert...
```suggestion
    assert '{search,add,edit,load}' in readout.out
```

happi/tests/test_cli.py
```diff
@@ -95,9 +124,214 @@ def test_search_z(happi_cfg):
...
+@pytest.mark.parametrize("from_user, expected_output", [(
+    # Test add item - succeeding
+    ['HappiItem', 'happi_name', 'device_class', ['arg1', 'arg2'],
+        {'name': 'my_name'}, True, 'docs', 'y'],
+    [
+        "Please select a container, or press enter for generic "
+        "Ophyd Device container",
+        "Enter value for name, default=None, "
+        "enforce=re.compile('[a-z][a-z\\\\_0-9]{2,78}$')",
+        "Enter value for device_class, default=None, enforce=<class 'str'>",
```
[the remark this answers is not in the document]
cristinasewell: Yeah I see what you mean - we would easily break the tests when modifying something related to the enforce.... 
Should I change it for all of them that use `enforce` then?

happi/tests/test_cli.py
cristinasewell: Changed this in 8d8698a - thank you

happi/tests/test_cli.py
```diff
@@ -95,9 +124,214 @@ def test_search_z(happi_cfg):
...
+    # we're going to search for z=3,7 name=TST_BASE_PIM2
+    # we should only get in return one entry, even though the z value found 2
+    res_cli = happi.cli.happi_cli(['--verbose', '--path', happi_cfg, 'search',
+                                   'name=TST_BASE_PIM2', 'z=3.0,7.0'])
+    assert [r.device for r in res] == [r.device for r in res_cli]
+
+
+@pytest.mark.parametrize("from_user, expected_output", [(
+    # Test add item - succeeding
+    ['HappiItem', 'happi_name', 'device_class', ['arg1', 'arg2'],
```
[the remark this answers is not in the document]
cristinasewell: Yeah that was a good idea, thank you. Added those in 8d8698a

happi/tests/test_cli.py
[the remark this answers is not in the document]
klauer: I think this is good enough that we can revisit it if/when this becomes an issue. It may be far down the line we decide to update those criteria, anyway.

happi/tests/test_cli.py
```diff
@@ -95,9 +122,200 @@ def test_search_z(happi_cfg):
...
+        "Enter value for documentation, default=None, enforce=<class 'str'>",
+        "Please confirm the following info is correct:",
+        "Adding device",
+        "Storing device HappiItem (name=happi_name) ...",
+        "Adding / Modifying information for happi_name ...",
+        "HappiItem HappiItem (name=happi_name) has been "
+        "succesfully added to the database"],
+    ),
+    # Test add item - aborting
+    pytest.param(['HappiItem', 'happi_name2', 'device_class', ['arg1', 'arg2'],
```
klauer, line 171: Good on the `pytest.param`, but you're still missing the `id='useful_identifier'` keyword argument.

happi/tests/test_cli.py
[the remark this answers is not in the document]
cristinasewell: Ok sounds good then, it will be pretty easy changing it to using regular expressions anyway, after we modify the `expected` strings like you mentioned above, we would then do something similar to this to assert them:
```
reg = re.compile(expected)
assert reg.match(message)
```

happi/tests/test_cli.py
cristinasewell: ohh i see hahah, let me add that in

happi/tests/test_cli.py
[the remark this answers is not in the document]
klauer: Yep - definitely!

happi/tests/test_cli.py
cristinasewell: Ok i think this is what you meant: 1568fc7 - but please let me know if I still did not get it 😄 - i won't mind, i don't think I have ever used the `ids` before  - it's good to know about them.
Also, please let me know if I should have provided something more generic for the IDs. 

I am able to now use the `-k id_value` to test a specific one which is cool

happi/tests/test_cli.py
```diff
@@ -79,6 +83,14 @@ def test_cli_version(capsys):
     assert happi.__file__ in readout.out
 
 
+def test_cli_no_argument(capsys):
+    happi.cli.happi_cli([])
+    readout = capsys.readouterr()
+    assert 'usage:' in readout.out
+    assert '[-h] [--path PATH] [--verbose] [--version]' in readout.out
+    assert '{search,add,edit,load}' in readout.out
```
ZLLentz, line 91: I think very specific reflexive tests like this are a maintenance hassle. What if argparse changes its output format slightly? What if we add an argument?

In cases like this I think less is more, e.g. just asserting that `usage` is in there would have been okay. If there's a need to make it specific, we should instead be getting the parser and calling `parser.print_usage()`, to make sure it matches what capsys sees. That way at least we don't have to modify this test for every cli interface update.

happi/tests/test_cli.py
```diff
@@ -87,6 +99,21 @@ def test_search(happi_cfg):
...
+    res = client.search_regex(name='TST_BASE_PIM2')
+    res_cli = happi.cli.happi_cli(['--verbose', '--path', happi_cfg, 'search',
+                                   'TST_BASE_PIM2'])
+    assert [r.device for r in res] == [r.device for r in res_cli]
+    # test duplicate search parameters
+    with caplog.at_level(logging.ERROR):
+        res_cli = happi.cli.happi_cli(['--verbose', '--path', happi_cfg,
+                                       'search', 'TST_BASE_PIM2',
+                                       'name=TST_BASE_PIM2'])
+        assert "Received duplicate search criteria" in caplog.text
```
ZLLentz, line 113: This is less of a problem than the one above since we control the output 100%, but in general I think it would be better to set up some return codes in `happi_cli` to check which branch we went down rather than relying on log messages.

This is more of a general thought, editing the return structure of the cli file to make the testing smoother doesn't have to be in scope for this PR.

happi/tests/test_cli.py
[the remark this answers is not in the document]
ZLLentz: I don't think we should be matching log or print output at all in this test suite, due to the volatility it introduces. Unit testing CLI is hard for this reason. Regex would be better, but even better would be breaking up the cli modules and testing their parts independently. I think that's the direction we should go if we revisit the test suite later.

happi/tests/test_cli.py
cristinasewell: Ok that sounds good, i'll update to only check for `usage` for now...

happi/tests/test_cli.py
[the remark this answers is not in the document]
cristinasewell: Yes I agree that refactoring the cli module to be able to test each command & their parts independently  would be a much better approach. Should I take these tests out from here?

happi/tests/test_cli.py
cristinasewell: Yeah I think this should be solved with refactoring the `cli.py` module to allow for that.....

happi/tests/test_cli.py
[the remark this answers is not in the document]
klauer: I think the coverage is worthwhile and your efforts here were useful. Regardless, I don't mind deferring to @ZLLentz's decision here.

happi/tests/test_cli.py
klauer: To be fair, in the issue, we did say "please help us write tests to get coverage" and not "refactor, break up cli.py, and make tests".

I think better communication between our groups from both sides could have saved you a bit of time here. We're always up for chatting on Slack, so open up a thread with us for future efforts and we can discuss ideas/concerns/thoughts/etc.

cristinasewell: Yeah it seems like overall I had the wrong approach with this PR, since pretty much I am checking against log messages for most of the tests that I added... maybe a better approach would have been to have an Issue that refactors the `cli` module first and then have an Issue for testing it...

happi/tests/test_cli.py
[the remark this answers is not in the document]
ZLLentz: I think we should merge this, and keep in mind the refactor if we revisit the test suite.

happi/tests/test_cli.py
ZLLentz: I think I didn't realize how awkward the whole cli.py was architected in the first place. That's probably my fault. It doesn't help that I'm getting repeatedly pulled into looking at other projects and am having trouble keeping up here...

klauer: No judgement intended, @ZLLentz, sorry if it sounded that way! Reiterating my usual offer, let me know if I can help with the other requests.

klauer:
As this is an improvement to the status quo, let's go forward with it. We can always revisit.  Thanks, @cristinasewell. 👍

happi/tests/test_cli.py
cristinasewell: @ZLLentz same here, I did not mean any judgement on anybody except myself haha, It was my bad, I think I should have talked to you on how we want to approach this PR, and that would have been  much better

PR #168, opened 2020-09-03
Happi audit

cristinasewell:
## Description

THIS IS STILL A WORK IN PROGRESS.....will update this when closer to be finished with it, for now I just need some advice/feedback before proceeding further

### DIFF
--- a/happi/audit.py
+++ b/happi/audit.py
@@ -5,8 +5,6 @@ from abc import ABC, abstractmethod
 import logging
 import os
 import sys
-import subprocess
-import re
 from happi.client import Client
 from happi.loader import import_class, create_arg
 from happi.containers import registry
@@ -72,8 +70,6 @@ class Audit(Command):
     def __init__(self):
         self.name = 'audit'
         self.help = "Inspect a database's entries"
-        self._all_devices = set()
-        self._all_items = []
         self.report_code = ReportCode.NO_CODE
         self._extras = False
         self._database_path = None
@@ -170,9 +166,9 @@ class Audit(Command):
         elif not container:
             logger.error('No container provided for %s', device)
             return self.report_code.MISSING
-        else:
-            # seems like the container has been validated
-            return self.report_code.SUCCESS
+
+        # seems like the container has been validated
+        return self.report_code.SUCCESS
 
     def validate_args(self, item):
         """
@@ -187,9 +183,9 @@ class Audit(Command):
         return dict((key, create_arg(item, val))
                     for key, val in item.kwargs.items())
 
-    def get_device_class(self, item):
+    def validate_import_class(self, item=None):
         """
-        Validates device_class field
+        Validate the device_class of an item
 
         Parameters
         ----------
@@ -203,106 +199,20 @@ class Audit(Command):
             logger.warning('Detected a None value for %s. '
                            'The device_class cannot be None', device)
             return self.report_code.MISSING
-        try:
-            mod, cls = device_class.rsplit('.', 1)
-        except Exception as e:
-            logger.warning('Wrong device name format: %s for %s, %s',
-                           device_class, device, e)
-            return self.report_code.INVALID
-        else:
-            packages = device_class.rsplit('.')
-            package = packages[0]
-            self._all_devices.add(package)
-            self._all_items.append(item)
-            return self._all_devices
-
-    def validate_import_class(self, item=None):
-        """
-        Validate the device_class of an item
-
-        Parameters
-        ----------
-        item: dict
-            An entry in the database
-        """
-        device_class = item.get('device_class')
-        mod = device_class
-        if not device_class:
-            return self.report_code.MISSING
         if '.' not in device_class:
             logger.warning('Device class invalid %s for item %s',
                            device_class, item)
             return self.report_code.INVALID
-        else:
-            mod = device_class.rsplit('.')[0]
-        device = item.get('name')
-        if mod in self._all_devices:
-            # has been tested agains PyPi and was found there
-            try:
-                import_class(device_class)
-                return self.report_code.SUCCESS
-            except ImportError as e:
-                logger.warning('For device: %s, %s. Either %s is '
-                               'misspelled or %s is not part of the '
-                               'python environment', device, e,
-                               device_class, device_class)
-                return (self.report_code.INVALID, self.report_code.MISSING)
-        elif mod and mod not in self._all_devices:
-            logger.warning('Provided wrong device/module name: %s. '
-                           'This module does not exist in PyPi', device_class)
-            return self.report_code.INVALID
 
-    def search_pip_package(self, package):
-        """
-        Checks to see if the package is on pypi
-
-        Parameters
-        -----------
-        package: str
-            Name of the package to check
-
-        Returns
-        --------
-            bool
-                To indicate if the package was found or not
-        """
-        process = None
         try:
-            process = subprocess.Popen('/bin/bash', stdin=subprocess.PIPE,
-                                       stdout=subprocess.PIPE,
-                                       encoding='utf-8')
-            arguments = ['pip', 'search', package]
-            command = ' '.join(arguments)
-
-            out, err = process.communicate(command)
-
-            # the output of 'pip search' will give all the fuzzy searches
-            # we only care if the exact package name is there
-            the_match = re.compile(package)
-            if re.match(the_match, out):
-                return True
-            else:
-                return False
+            import_class(device_class)
+            return self.report_code.SUCCESS
         except Exception as e:
-            logger.warning(e)
-            if process:
-                process.kill()
-
-    def check_device_in_pypi(self):
-        if self._all_devices:
-            temp_set = self._all_devices.copy()
-            for package in self._all_devices:
-                is_package_found = self.search_pip_package(package)
-                if is_package_found:
-                    logger.info('%s was found in PyPi', package)
-                else:
-                    # maybe a wrong device name was provided because
-                    # it does not exit
-                    logger.info('%s does not exit in PyPi', package)
-                    temp_set.remove(package)
-            # remove the missing package from the set
-            self._all_devices = temp_set
-            return self._all_devices
+            logger.warning('For device: %s, %s. Either %s is '
+                           'misspelled or %s is not part of the '
+                           'python environment', device, e,
+                           device_class, device_class)
+            return (self.report_code.INVALID)
 
     def validate_enforce(self, item):
         """
@@ -368,9 +278,8 @@ class Audit(Command):
             logger.warning('Device %s has extra attributes %s',
                            item.name, diff)
             return self.report_code.EXTRAS
-        else:
-            # no extra attributes found
-            return self.report_code.SUCCESS
+
+        return self.report_code.SUCCESS
 
     def parse_database(self, database_path):
         """
@@ -416,16 +325,11 @@ class Audit(Command):
         for item in client.backend.all_devices:
             it = client.find_document(**item)
             self.validate_enforce(it)
-            self.get_device_class(it)
 
-        # make sure to call this function after get_device_class
-        print_report_message('VALIDATING DEVICE MODULE EXISTS IN PYPI')
-        self.check_device_in_pypi()
         # validate import_class
         print_report_message('VALIDATING DEVICE CLASS')
-        if self._all_items:
-            for item in self._all_items:
-                self.validate_import_class(item)
+        for item in items:
+            self.validate_import_class(item)
 
         print_report_message('VALIDATING CONTAINER')
         for item in client.backend.all_devices:
--- a/happi/audit.py
+++ b/happi/audit.py
@@ -6,14 +6,17 @@ Happpi configuration file using the Client.find_config method,
 otherwiwe it will use the path specified after --file
 """
 from configparser import ConfigParser
-import json
 import logging
 import os
 import sys
 
-import happi
-
 from happi.client import Client
+
+# For back-compat to <py3.7
+try:
+    from re import Pattern
+except ImportError:
+    from re import _pattern_type as Pattern
 
 
 logger = logging.getLogger(__name__)
@@ -22,7 +25,7 @@ logger = logging.getLogger(__name__)
 class Command(object):
     """
     Bluprint for the command class
-    This will be useful if all the commands are inhering from it
+    This will be useful if all the commands are inheriting from it
     """
     def __init__(self, name, summary):
         self.name = name
@@ -38,6 +41,9 @@ class Command(object):
 
 
 class Audit(Command):
+    """
+    Audits the database for valid entries.
+    """
     def __init__(self):
         self.name = 'audit'
         self.help = "Inspect a database's entries"
@@ -84,26 +90,123 @@ class Audit(Command):
     def validate_file(self, file_path):
         """
         Tests whether a path is a regular file
+
+        Parameters
+        -----------
+        file_path: str
+            File path to be validate
+
+        Returns
+        -------
+        bool
+            To indicate whether the file is a valid regular file
         """
         return os.path.isfile(file_path)
 
+    # TODO this is probably not needed, it assumes only json database...
+    # but keep it here for reference for now
+    # def parse_database(self, database_path):
+    #     """
+    #     Goes through an entire database and parses
+    #     all the entries
+
+    #     Parameters
+    #     -----------
+    #     database_path: str
+    #         Path to the database to be validated
+    #     """
+    #     # check for an empty database???
+    #     # check for an invalid file
+    #     # we are assuming
+    #     registry = happi.containers.registry
+    #     with open(database_path) as f:
+    #         data = json.loads(f.read())
+    #         for key, value in data.items():
+    #             # here i have to also check if the type is
+    #             # correct, and if it has a type???
+    #             try:
+    #                 container = registry[value.get('type')]
+    #                 logger.info('Container %s', container)
+    #             except Exception as ex:
+    #                 logger.exception('Something went wrong with '
+    #                                  'getting the type of the item: %s', ex)
+    #             self.validate_item(database_path, container, value)
+
+    def validate_mandatory_info(self, item):
+        for info in item.entry_info:
+            # if it is not optional, it should probably never be None
+            if not info.optional and item.get(info.key) is None:
+                logger.info('Entry %s must not have None for '
+                            'a mandatory entry: %s', item, info)
+
+    def validate_enforce(self, item):
+        for info in item.entry_info:
+            value = item.get(info.key)
+            if isinstance(info.enforce, type):
+                if info.enforce == str:
+                    # check the values that are enforced to be strings
+                    # TODO - not working for all items, if i have a number
+                    # inserted into an attribute where it should be string
+                    # this will still consider it as string.... not what i want
+                    if (not isinstance(value, str) and (value is not None) and
+                       info.default is not None):
+                        logger.info('The %s did not match the enforced type '
+                                    '%s for the entry: %s. Current value: %s',
+                                    info, info.enforce, item, value)
+                if info.enforce == float:
+                    # check the values that are enforced to be floats
+                    # TODO - this is not going to work here, because
+                    # even if i have something like '343.43' for z value
+                    # where it has quotes and inserted as string basically
+                    # this will still work and not consider it a bad entry
+                    print(f'The ones that are floats: {value}')
+                    if (not isinstance(value, float)):
+                        logger.info('The %s did not match the enforced type '
+                                    '%s for the entry: %s. Current value: %s',
+                                    info, info.enforce, item, value)
+                if info.enforce == list:
+                    # check the values that are enforced to be lists
+                    pass
+                if info.enforce == dict:
+                    pass
+                    # check the values that are enforced to be dictionaries
+                if info.enforce == bool:
+                    pass
+                    # check the values that are enforced to be boolean
+
+            # check the values that are enforced with regex
+            elif isinstance(info.enforce, Pattern):
+                if not info.enforce.match(str(value)):
+                    logger.info('The %s did not match the enforced pattern '
+                                '%s for the entry: %s. Current value: %s',
+                                info, info.enforce, item, value)
+
     def parse_database(self, database_path):
         """
-        Goes through an entire database and validates
-        all the entries there
-        """
-        registry = happi.containers.registry
-        with open(database_path) as f:
-            data = json.loads(f.read())
-            for key, value in data.items():
-                # here i have to also check if the type is
-                # correct, and if it has a type???
-                container = registry[value.get('type')]
-                logger.debug('Container %s ', container)
+        Validates if an entry is a valid entry in the database
 
-    def validate_item(self, container, item_list):
-        try:
-            #  to validate
-            pass
-        except Exception as e:
-            logger.error(e)
+        Parameters
+        ----------
+        database_path: str
+            Path to the database to be validated
+
+        """
+        it = None
+        client = Client(path=database_path)
+        for item in client.backend.all_devices:
+            items = client.find_document(**item)
+            try:
+                # try to instantiate so you can have access to entry_info
+                it = client.create_device(items['type'], **items)
+                self.validate_mandatory_info(it)
+                self.validate_enforce(it)
+            except Exception:
+                # TODO handle these guys differently...
+                logger.info('Could not create a device.....')
+
+        # TODO if using the all_items I run
+        # into exceptions and can't look at all the devices
+        # items = client.all_items
+        # for item in items:
+        #   self.validate_mandatory_info(item)
+        #   self.validate_enforce(item)
--- a/happi/audit.py
+++ b/happi/audit.py
@@ -4,7 +4,6 @@ This module defines the ``happi audit`` command line utility
 from abc import ABC, abstractmethod
 import logging
 import os
-import sys
 from happi.client import Client
 from happi.loader import import_class, create_arg
 from happi.containers import registry
@@ -93,27 +92,17 @@ class Audit(Command):
         self._extras = args.extras
 
         if args.file is not None:
-            process_database(database_path=args.file, extras=self._extras)
+            if validate_file(args.file):
+                process_database(database_path=args.file, extras=self._extras)
+            else:
+                logger.error('The database %s file path '
+                             'could not be validated', args.file)
+                return
         else:
             """
             Validate the database defined in happi.cfg file
             """
-            path_to_config = Client.find_config()
-
-            if path_to_config:
-                logger.info('Using Client cfg path %s ', path_to_config)
-                db_kwargs = Client.parse_config_file(path_to_config)
-                self._database_path = db_kwargs.get('path')
-                if self._database_path:
-                    process_database(database_path=self._database_path,
-                                     extras=self._extras)
-                else:
-                    logger.error('Could not get the database path '
-                                 'from the configuration file')
-                    return
-            else:
-                logger.error('Could not find the Happi Configuration file')
-                sys.exit(1)
+            process_database(extras=self._extras)
 
 
 def process_database(database_path=None, extras=False):
@@ -122,16 +111,10 @@ def process_database(database_path=None, extras=False):
     If extras is True, it will check for extra attributes only, otherwise
     it will just proceed with parsing and validating the database call.
     """
-    if validate_file(database_path):
-        logger.info('Using database file at %s ', database_path)
-        if extras:
-            check_extra_attributes(database_path)
-        else:
-            parse_database(database_path)
+    if extras:
+        check_extra_attributes(database_path)
     else:
-        logger.error('The database %s file path '
-                     'could not be validated', database_path)
-        sys.exit(1)
+        audit_information_from_client(database_path)
 
 
 def validate_file(file_path):
@@ -248,7 +231,7 @@ def validate_enforce(item):
         return ReportCode.MISSING
 
 
-def check_extra_attributes(database_path):
+def check_extra_attributes(database_path=None):
     """
     Checks the entries that have extra attributes
 
@@ -258,7 +241,10 @@ def check_extra_attributes(database_path):
         Path to the database to be validated
 
     """
-    client = Client(path=database_path)
+    if database_path is not None:
+        client = Client(path=database_path)
+    else:
+        client = Client.from_config()
     items = client.all_items
 
     print_report_message('EXTRA ATTRIBUTES')
@@ -291,7 +277,7 @@ def validate_extra_attributes(item):
     return ReportCode.SUCCESS
 
 
-def parse_database(database_path):
+def audit_information_from_client(database_path=None):
     """
     Validates if an entry is a valid entry in the database
 
@@ -301,7 +287,11 @@ def parse_database(database_path):
         Path to the database to be validated
 
     """
-    client = Client(path=database_path)
+    if database_path is not None:
+        client = Client(path=database_path)
+    else:
+        client = Client.from_config()
+
     items = client.all_items
 
     print_report_message('VALIDATING ENTRIES')
--- a/happi/audit.py
+++ b/happi/audit.py
@@ -15,8 +15,19 @@ import re
 from happi.client import Client
 from happi.loader import import_class, fill_template
 from happi.containers import registry
+from enum import Enum
 
 logger = logging.getLogger(__name__)
+
+
+class ReportCode(Enum):
+    """
+    Class to define report codes
+    """
+    SUCCESS = 0
+    INVALID = 1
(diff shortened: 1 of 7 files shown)


cristinasewell:
@ZLLentz  can you please give me some help/feedback on this one before proceeding further??
It seems like I am going a bit in circles when validating certain things...
I tried to use functions that are already defined in the loader so I don't repeat code, but still, some of the things are not fully validated I think....


I also found a couple of bugs that i fixed here, the major one was introduced by me in the search for cli.py when searching both with regex and range and one of them did not validate anything... i fixed it here because i found it here when testing...

happi/audit.py
```diff
@@ -158,49 +158,21 @@ def validate_device_class(self, item):
                 except ImportError as ex:
                     logger.warning(ex)
 
-    def validate_enforce(self, entry, item):
-        for (key, value), info in zip(item.items(), entry.entry_info):
-            # print(f'item_______ {value}')
-            # print(type(value))
-            # print(info.enforce)
-            pass
+    def validate_extra_attributes(self, item):
```
cristinasewell, line 161: This function will need to be updated, just realized that this will be valid only for `OphydItems` or `HappiItems`.... not necessarily for some other containers.

happi/audit.py
```diff
@@ -0,0 +1,218 @@
...
+        ----------
+        database_path: str
+            Path to the database to be validated
+
+        """
+        client = Client(path=database_path)
+        items = client.all_items
+
+        self.print_report_message('VALIDATING ENTRIES')
+        client.validate()
```
cristinasewell, line 196: Also I might have to make my own `validate()` here - as i kind of started previously, but i am not sure, for now when i do this it gives me warning like these:
`Failed to validate cxi_dg1_slits because Missing mandatory information (device_class, location_group, functional_group) for Slits` - i might have to have extra checks here or something else, these devices seem to have a device_class already...

ZLLentz:
Sorry for slow responses here.

I'd like to reiterate that I think an auditing command is a hard thing to set up, so feel free to ask for input at any time.

I'll look through this draft now

cristinasewell:
Thanks a lot @ZLLentz :)

happi/audit.py
ZLLentz: I don't see any particular assumptions in this method that wouldn't work for arbitrary container definitions.

happi/audit.py
ZLLentz: Also that repo is super confusing right now: `master` branch is an old snapshot that should probably fail the audit, while the `deploy` branch is a new setup that should probably mostly pass.

happi/audit.py
```diff
@@ -0,0 +1,218 @@
...
+        for item in items:
+            self.validate_args(item)
+            self.validate_kwargs(item)
+            self.validate_extra_attributes(item)
+            pass
+
+        # TODO: what to do here???....
+        # for entr, item in zip(items, client.backend.all_devices):
+        #     #self.validate_enforce(entr, item)
+        #     pass
```
ZLLentz, line 208: I think https://github.com/pcdshub/happi/blob/c77ed7d28c74d946800afad5a60ced5f9e317500/happi/item.py#L82 is the key function to use here, to check that the values saved in the item match their respective EntryInfo

happi/cli.py
```diff
@@ -208,8 +219,8 @@ def happi_cli(args):
                     info.enforce_value(item_value)
                     valid_value = True
                     kwargs[info.key] = item_value
-                except Exception:
-                    logger.info(f'Invalid value {item_value}')
+                except Exception as e:
+                    logger.info('Invalid value %s, %s', item_value, e)
```
ZLLentz, line 223: another way to log exceptions that you may want to try is:
```
logger.info(text, exc_info=True)
```
This logs the full traceback, which sometimes is what you want.

happi/audit.py
cristinasewell: Yeah just realized that this morning :).
Is there a good way to differentiate between a module that we know it is valid but not in the python environment, other than specifically checking for `pcdsdevices`? - I want to be able to know when it is invalid/misspelled vs valid but not in the path?

ZLLentz:
I think this coming along pretty nicely- based on your messages, I assumed you weren't this far.

happi/audit.py
cristinasewell: Ok thank you, i need to revisit this one, it might have been related to the fact that I did not have `pcdsdevices` in my path...

happi/audit.py
cristinasewell: Yes I see what you mean, I am testing against `enforce_value` already - i like it, and it works great as long as we only add to the database using `happi` :) 
If for example I have a number instead of  a string or even a numerical string instead of float - this will still work, but the loading will not work, and I wanted to see if I can catch that - but if I already initialize an EntryInfo, it is too late at that point to check for that I think, it has to be happening before even initializing it.....

happi/cli.py
cristinasewell: This is great, thank you!

happi/audit.py
ZLLentz: I'm not sure there are great tools for a generic python environment to check if a module is "real and missing" vs "not a real thing" (misspelled). Our best effort here could be to check pypi or conda, which should typically be available. I'm not sure if those have easily accessible python bindings or if we need to shell out e.g. `pip search pcdsdevices`. It also might be possible to get a list of installed modules and try to do a fuzzy match, e.g. "you specified the pcdsevices module, did you mean pcdsdevices?"

happi/audit.py
ZLLentz: I'm not sure I fully understand. Are you saying that if someone manually edits the json, an enforce_value can pass, despite being in a form that fails in the loader later?

happi/audit.py
cristinasewell: Ohh that's a great idea, i should look into that, i did not think of checking with pypi and conda - i hope it is possible and not too bad to use :) - thank you for the suggestions here

happi/audit.py
cristinasewell: Maybe I got it wrong here, i need to revisit this one, but for example if i try to add this item:  **Please note that i have a string for the `z` value**

```
device_info = {'name': 'alias3',
            'z': '400',
            '_id': 'alias3',
            'prefix': 'BASE:PV',
            'beamline': 'LCLS',
            'type': 'OphydItem',
            'device_class': 'types.SimpleNamespace',
            'args': list(),
            'kwargs': {'hi': 'oh hello'},
            'location_group': 'LOC',
            'functional_group': 'FUNC',
            }
device = OphydItem(**device_info)
client.add_device(device)

```
This will add fine, will also validate fine, and even load fine - but we might have problems when we look for this z value and try to search it... 

One other example I can think of is this one: **Note the beamline == True** - this will convert fine to a string so it won't give us issues saving it, or even loading it - but technically it is not a valid entry??

```
device_info = {'name': 'alias4',
            'z': '400',
            '_id': 'alias4',
            'prefix': 'BASE:PV',
            'beamline': True,
            'type': 'OphydItem',
            'device_class': 'types.SimpleNamespace',
            'args': list(),
            'kwargs': {'hi': 'oh hello'},
            'location_group': 'LOC',
            'functional_group': 'FUNC',
            }
```

Maybe I am being a bit too much here though hahah - please let me know :)

ZLLentz: It seems like `happi` is being very permissive for a lot of these fields: as long as you can cast it to the required type, it is ok as far as `happi` is concerned. I haven't thought for long enough if this is something that should be supported.

If this is causing issues with `search` then we clearly have to intervene a bit... I'm not sure the best resolution. It sounds just as much like a `search` bug as a validation error though.

cristinasewell: I agree, I think as long as we can cast it to the correct type we should be fine - which we already do, and also if we have problems searching, this should be part of the searching logic to handle. Thank you.

ZLLentz:
Is this ready for another review?

cristinasewell:
@ZLLentz I would really like some feedback for this if you have time. I have not worked on it since last week, and will continue working on it on Monday, but some feedback by then would be very useful. 

Thanks a lot :)

ZLLentz:
I will give it a review now and touch base with you on Monday then :+1:

happi/audit.py
```diff
@@ -0,0 +1,423 @@
+"""
+This module defines the ``happi audit`` command line utility
+
+If the --file option is not provided, then it will take the path from
+Happpi configuration file using the Client.find_config method,
```
ZLLentz, line 5:
```suggestion
Happi configuration file using the Client.find_config method,
```

happi/audit.py
```diff
@@ -0,0 +1,423 @@
...
+otherwiwe it will use the path specified after --file
```
ZLLentz, line 6:
```suggestion
otherwise it will use the path specified after --file
```

happi/audit.py
```diff
@@ -0,0 +1,423 @@
...
+    INVALID = 1
+    MISSING = 2
+    EXTRAS = 3
+    NO_CODE = 9
+
+
+class Command(object, metaclass=ABCMeta):
+    """
+    Bluprint for the command class
+    This will be useful if all the commands are inheriting from it
```
ZLLentz, line 37: This is a good target for a future refactor

happi/audit.py
```diff
@@ -0,0 +1,423 @@
...
+            Validate the database defined in happi.cfg file
+            """
+            path_to_config = Client.find_config()
+
+            if path_to_config:
+                logger.info('Using Client cfg path %s ', path_to_config)
+                config = ConfigParser()
+                config.read(path_to_config)
+                try:
+                    database_path = config.get('DEFAULT', 'path')
```
ZLLentz, line 115: I'm really surprised we don't have a centralized code path for going from config to database file

happi/audit.py
```diff
@@ -0,0 +1,423 @@
...
+                except Exception as e:
+                    logger.error('Error when trying '
+                                 'to get database path %s', e)
+                    return
+                if self.validate_file(database_path):
+                    logger.info('Using database file at %s ', database_path)
+                    if args.extras:
+                        self.check_extra_attributes(database_path)
+                    else:
+                        self.parse_database(database_path)
```
ZLLentz, line 125: There's just a little bit of repetition here with the lines above that could be consolidated into a function or method

happi/audit.py
```diff
@@ -0,0 +1,423 @@
...
+                else:
+                    logger.error('The database %s file path '
+                                 'could not be validated', database_path)
+                    sys.exit(1)
+            else:
+                logger.error('Could not find the Happi Configuration file')
+                sys.exit(1)
+
+    def validate_file(self, file_path):
```
ZLLentz, line 134: Stuff like this is easy to miss, glad you included it

happi/audit.py
```diff
@@ -0,0 +1,423 @@
...
+import re
+from happi.client import Client
+from happi.loader import import_class, fill_template
+from happi.containers import registry
+from enum import Enum
+
+logger = logging.getLogger(__name__)
+
+
+class ReportCode(Enum):
```
ZLLentz, line 23: I think using a report code enum was a good decision

ZLLentz:
I'm pretty happy with this. After some of the beamline python demands quiet down I'll give it a spin to have further feedback.

happi/audit.py
cristinasewell: We do have this handled in the Client, `from_config(cls, cfg=None)` but it just does extra stuff that I don't need - but maybe I overcomplicated this here, I might just be able to use the `from_config` method, or if not, I could pull the config to database file portion out and place it in its own little function...

happi/audit.py
cristinasewell: Ohh thanks for pointing this out, fixed it in b183eeb.

happi/audit.py
cristinasewell: thank you, fixed it in b183eeb

happi/audit.py
cristinasewell: fixed it in b183eeb

happi/audit.py
```diff
@@ -0,0 +1,424 @@
+"""
+This module defines the ``happi audit`` command line utility
+
+If the --file option is not provided, then it will take the path from
```
klauer, line 4: I think this should be explained in the `--file` argument and is unnecessary up here.

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+    Audits the database for valid entries.
+    """
+    def __init__(self):
+        self.name = 'audit'
+        self.help = "Inspect a database's entries"
+        self._all_devices = set()
+        self._all_items = []
+        # 9 - NO_CODE
+        self.report_code = ReportCode(9)
```
klauer, line 79:
```suggestion
        self.report_code = ReportCode.NO_CODE
```

When dealing with these enums - you should _never_ have to deal with the raw integer value, unless it's coming over the wire or something.

In this case, the values truly don't even matter: consider `enum.auto()` for the values and making the enum `__repr__` look something like `return self.name` to get a pretty repr.

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+    Class to define report codes
+    """
+    SUCCESS = 0
+    INVALID = 1
+    MISSING = 2
+    EXTRAS = 3
+    NO_CODE = 9
+
+
+class Command(object, metaclass=ABCMeta):
```
klauer, line 34:
```suggestion
class Command(abc.ABC):
```

Metaclasses are confusing - the less the better: https://docs.python.org/3.8/library/abc.html#abc.ABC

Separately, I'm clearly biased but I've had good luck with the following pattern: https://github.com/slaclab/pytmc/blob/master/pytmc/bin/pytmc.py#L27

Have a top-level command, let the sub-commands build their parsers and return them to you, associate a function and pass the parsed args in as keywords. 🤷‍♂️ 

I do like that you get a registry of subcommands for free with this `Command` class, though.

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+    Bluprint for the command class
```
klauer, line 36:
```suggestion
    Blueprint for the command class
```

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+    This will be useful if all the commands are inheriting from it
+    """
+
+    def __init__(self, name, summary):
+        self.name = name
```
klauer, line 41: Functionality should be removed in the ABC

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+        cmd_parser.add_argument(
+            "--extras", action='store_true', default=False,
+            help='Check specifically for extra attributes'
+        )
+
+    def run(self, args):
+        """
+        Validate the database passed in with --file option
+        """
+        if args.file is not None:
```
klauer, line 94: While it fits in with your `Command` class and it makes sense in this context, consider the following:

I want to use a Python script to validate 30 different databases without resorting to shelling out to `happi audit ...`.
How do I do that with this code? Instantiate this `Audit()` class, fake out the `parse_args()`, and then call `run()`?

The functionality of the class should be entirely broken out into separate, independently usable functions, not relying on the state held in the class. Additionally, the CLI interface to these functions should be relatively small.

Please let me know if you need further details here.

happi/audit.py
klauer: This should be refactored around `from_config`, as you noted. I don't see a good reason to duplicate its work in a slightly different way?

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+                else:
+                    self.process_args(database_path, args)
+            else:
+                logger.error('Could not find the Happi Configuration file')
+                sys.exit(1)
+
+    def process_args(self, database_path, args):
+        """
+        Checks to see if a valid path to database was provided.
+        If --extras is provided, will call for check xtra attributes, otherwise
```
klauer, line 121:
```suggestion
        If --extras is provided, will check extra attributes, otherwise
```

Additionally, `--extras` is a CLI argument, not a Python [keyword] argument. These should be functions with Python arguments, as I mentioned in the comment above.

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+        Returns
+        -------
+        bool
+            To indicate whether the file is a valid regular file
+        """
+        return os.path.isfile(file_path)
+
+    def validate_container(self, item):
+        """
+        Validates container definition
```
klauer, line 153: `item` should be in the `Parameters` docstring here and below

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+    def validate_kwargs(self, item):
+        """
+        Validates the kwargs of an item
+        """
+        return dict((key, self.create_arg(item, val))
+                    for key, val in item.kwargs.items())
+
+    def create_arg(self, item, arg):
+        """
+        Function borrowed from loader to create
```
klauer, line 183: Please pull this function out in the loader so it can be reused rather than duplicating it here.

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+        Validates device_class field
+        """
+        device_class = item.get('device_class')
+        device = item.get('name')
+
+        if not device_class:
+            logger.warning('Detected a None vlaue for %s. '
+                           'The device_class cannot be None', device)
+            return self.report_code.MISSING
+        else:
```
klauer, line 201: A stylistic thing, but for good practice: you don't need the `else:` block here as the above `if` has an unconditional `return`. That reduces the indentation significantly.

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+            try:
+                mod, cls = device_class.rsplit('.', 1)
+            except (Exception, ValueError) as e:
```
klauer, line 204:
```suggestion
            except Exception as e:
```
Exception will catch pretty much everything including `ValueError`

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+            try:
+                import_class(device_class)
+                return self.report_code.SUCCESS
+            except ImportError as e:
+                logger.warning('For device: %s, %s. Either %s is '
+                               'misspelled or %s is not part of the '
+                               'python environment', device, e,
+                               device_class, device_class)
+                return (self.report_code.INVALID, self.report_code.MISSING)
+        elif mod and mod not in self._all_devices:
```
klauer, line 241: Is the `elif` necessary here? Isn't the only alternative it being invalid?

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+            logger.warning('Provided wrong device/module name: %s. '
+                           'This module does not exist in PyPi', device_class)
+            return self.report_code.INVALID
+
+    def search_pip_package(self, package):
```
klauer, line 246: This is a cool idea, but I don't think it has a place in this audit script. It should just be up to the user to figure out why it's bad if it doesn't import.

It's a bit heavy, requires an internet connection, spawns new processes and slows down the audit process, etc.
Oh, and sometimes the import name of a package and its package name are two different things, even.

Edit: Reading below, it looks like @ZLLentz requested this. I think we should discuss this further.

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+                else:
+                    # maybe a wrong device name was provided because
+                    # it does not exit
+                    logger.info('%s does not exit in PyPi', package)
+                    temp_set.remove(package)
+            # remove the missing package from the set
+            self._all_devices = temp_set
+            return self._all_devices
+
+    def get_value(self, info, item):
```
klauer, line 298: `HappiItem` is a mapping, so it can be dictified `dict(item)[info.key]`

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+        for key, value in item.items():
+            if key == info.key:
+                return value
+
+    def validate_enforce(self, item):
+        """
+        Validates using enforce_value() from EntryInfo class
+        If the attributes are malformed the entry = contaienr(**item)
```
klauer, line 306:
```suggestion
        If the attributes are malformed the entry = container(**item)
```

happi/audit.py
```diff
@@ -0,0 +1,424 @@
...
+                    info.enforce_value(self.get_value(info, item))
+                    return self.report_code.SUCCESS
+                except Exception as e:
+                    logger.info('Invalid value %s, %s', info, e)
+                    return self.report_code.INVALID
+        else:
+            logger.warning('Item %s is missing the container', name)
+            return self.report_code.MISSING
+
+    def print_report_message(self, message):
```
klauer, line 324: A clue something doesn't belong as a method: this doesn't use `self` anywhere.

klauer (changes requested):
I didn't get to looking at the tests, but I think there's a fair amount of work that needs to be done to get this merged.  Please don't take offense - I think this can be a good learning experience for you.

happi/audit.py
cristinasewell: Yeah it works fine, but it takes some time to check, so it does slow down the audit process. Also, because it requires internet it is not ideal - I agree with that. 
I think you make a good point that we can leave it up to the user to figure out why it's bad if it does not import - what do you think @ZLLentz ?

happi/audit.py
cristinasewell: I agree - moved it in 1f87a91

happi/audit.py
cristinasewell: Ohh wow perfect, I did not know about the `auto()` option in Python - this is cool and exactly what I needed here, thanks a lot Ken - fixed it in 1f87a91, I hope is what you meant.

happi/audit.py
cristinasewell: Got it in 1f87a91, thank you

happi/audit.py
cristinasewell: Ok, took the constructor out.

happi/audit.py
cristinasewell: Ok sounds good, I pulled this function out in 4fabfae, but still left it in the loader - is this what you meant?

happi/audit.py
cristinasewell: Yep that's so right - fixed it in 1f87a91. Thank you

happi/audit.py
cristinasewell: Ok, added this in 1f87a91

happi/audit.py
cristinasewell: This might not be necessary here if I don't have to deal with missing containers from pypi.... 
Also if I don't have to deal with searching packages in pypi I will probably be able to do a bit more cleanup here too

happi/audit.py
cristinasewell: thanks a lot for this one.... haha, it bugged me that I could not find a one-liner for it

happi/audit.py
cristinasewell: got it in 1f87a91 - thank you

happi/audit.py
cristinasewell: ok I took it outside the class.... thanks :)

happi/audit.py
cristinasewell: Yeah I see. I moved the `parsing configuration file` logic from `from_config`  into a separate `static method` in the `client` - in 75cbcd7. I am not sure if that was the right approach - but all I need is the `file_path` for the database, and `from_config` uses that but does not return me the path - what do you think?

happi/audit.py
cristinasewell: Ohh I did not consider that...... 
I have implemented some changes related to this in 7cf8a0b, but I need to revisit it..... (especially if I implement your suggestion above https://github.com/slaclab/pytmc/blob/master/pytmc/bin/pytmc.py#L27  )
`run` and `add_args` should be the only ones dealing with `cli` now.... I will get back to you with this one. 

Thanks a lot for pointing this out..

cristinasewell:
> I didn't get to looking at the tests, but I think there's a fair amount of work that needs to be done to get this merged. Please don't take offense - I think this can be a good learning experience for you.

Thanks a lot for taking the time to review and to give me great feedback Ken - I appreciate it ;)

happi/audit.py
ZLLentz: It's not fully necessary to do the pip search. Maybe we were a little "too helpful" here in ways where we make the script cumbersome to run.

happi/audit.py
klauer: Looks good overall, but isn't the repr you implemented exactly as the default one provided by `enum.Enum`?

https://github.com/pcdshub/happi/commit/1f87a9198276c31eb8e7fb68e79c277ff1c29678#diff-e967c494e6919641a643a09b9e94bddcR35-R36

happi/audit.py
klauer: It could be an option, given that you've done the work here @cristinasewell. Making it optional will further complicate the already pretty complicated logic here, though. Is pypi searching something we could perhaps shelve for the next audit version?

klauer:
OK, I think you've got most of the smaller stuff. 

The bigger tasks remaining is moving all of the methods from the `Audit` command out into self-sufficient functions in the module. All of the validation methods can be stateless - or at least can be refactored to be stateless.

I don't quite understand why you need to deal with the database filename instead of using the abstractions that `happi.Client` provides for you. `audit_information_from_client(happi.Client.from_config())` -> `for item in client.all_items: validate_item(item)` naively seems all that should be necessary to me.

happi/audit.py
cristinasewell: Probably, although when I do this: `print(repr(self.report_code.EXTRAS))` without the `__repr__()` I get the value as well, something like this: 
<ReportCode.EXTRAS: 4>

And  this: `return '<%s.%s>' % (self.__class__.__name__, self.name)` hides the value: 
<ReportCode.EXTRAS>

Regardless, I probably don't need to define it here right?

happi/audit.py
cristinasewell: I am a fan of removing this if that's ok :)

cristinasewell:
> OK, I think you've got most of the smaller stuff.
> 
> The bigger tasks remaining is moving all of the methods from the `Audit` command out into self-sufficient functions in the module. All of the validation methods can be stateless - or at least can be refactored to be stateless.
> 
> I don't quite understand why you need to deal with the database filename instead of using the abstractions that `happi.Client` provides for you. `audit_information_from_client(happi.Client.from_config())` -> `for item in client.all_items: validate_item(item)` naively seems all that should be necessary to me.

There were a few things (none of them really justify why I am using the database file though haha ): 
- I think I initially wanted to approach the audit differently and use a json schema to validate, and that's how I ended up using the filename of the database instead of using `happi.Client.from_config()`. 
- Also, I gave the option to the `happi audit` to provide a filename: `happi audit  --file db.json`  (and if I use the database in both cases it make the logic simpler)- but this should probably be eliminated, since it only really works with json database.....  
- Also, when I do `client.all_items` - it already skips the "malformed" entries (and can't really see what is really wrong with them exactly),  and I think I was trying to avoid that by going around `client.all_items` and go straight to the database entries - but obviously I am not avoiding that anyway.... 

I will change it to use `happi.Client.from_config()`. 
Should I get rid of the `happi audit --file ` option too?  - it is probably not necessary as well?

klauer:
> I think I initially wanted to approach the audit differently and use a json schema to validate, and that's how I ended up using the filename of the database instead of using happi.Client.from_config().

Makes sense - but yeah as you know, we don't want to restrict ourselves to the JSON backend (even if it is all we're using at the moment).

> Also, I gave the option to the happi audit to provide a filename: happi audit --file db.json (and if I use the database in both cases it make the logic simpler)- but this should probably be eliminated, since it only really works with json database.....

I think `audit --config config_filename` could be fine. It's up to the backend to use the config file (which then may point to the database host/filename/URL/etc). 

> Also, when I do client.all_items - it already skips the "malformed" entries and I think I was trying to avoid that by going around client.all_items and go straight to the database entries - but obviously I am not avoiding that anyway....

Hmm - which scenarios would invalid items not show up in `all_items`?

Generally speaking, almost all of our interaction should be with _just_ the client here. If we need something from the backend that can't be accessed from the client itself, it should be added to all the backends and then the client should have a method that allows you to access the information from any backend uniformly. Perhaps retrieving invalid items is something that needs thought here.

cristinasewell:
> > I think I initially wanted to approach the audit differently and use a json schema to validate, and that's how I ended up using the filename of the database instead of using happi.Client.from_config().
> 
> Makes sense - but yeah as you know, we don't want to restrict ourselves to the JSON backend (even if it is all we're using at the moment).
> > > yep - i'll get rid of that  

> > Also, I gave the option to the happi audit to provide a filename: happi audit --file db.json (and if I use the database in both cases it make the logic simpler)- but this should probably be eliminated, since it only really works with json database.....
> 
> I think `audit --config config_filename` could be fine. It's up to the backend to use the config file (which then may point to the database host/filename/URL/etc).
> >>ohh i see, this would make way more sense to do... do you think it is necessary though, should I do it, or is it going to be one of those features that people won't even know about because they'll never need to use it :) ?

> > Also, when I do client.all_items - it already skips the "malformed" entries and I think I was trying to avoid that by going around client.all_items and go straight to the database entries - but obviously I am not avoiding that anyway....
> 
> Hmm - which scenarios would invalid items not show up in `all_items`?

>>>If there is something wrong with the `device_class` or an invalid `name` - maybe others too... 
I think it is related to the `_get_search_results` method, this `result = wrap_cls(client=self, device=self.find_device(**info))` fails if something is wrong (which I guess it is fine because I am catching the exception from this one?)

> Generally speaking, almost all of our interaction should be with _just_ the client here. If we need something from the backend that can't be accessed from the client itself, it should be added to all the backends and then the client should have a method that allows you to access the information from any backend uniformly. Perhaps retrieving invalid items is something that needs thought here.
👍

klauer:
> `audit --config`

I don't think the `--config` is necessarily a requirement as setting `HAPPI_CFG` could be sufficient. Since `Client.from_config()` lets you pass in a configuration file, it seems easy enough to support it here though if it seems useful.

> I think it is related to the `_get_search_results` method, this `result = wrap_cls(client=self, device=self.find_device(**info))` fails if something is wrong (which I guess it is fine because I am catching the exception from this one?)

Hmm, I see. Restating what we're after here: auditing needs to know about both items that can load (to check their information) and items that can't be loaded (to see what the problem is). For the latter, `_get_search_results` is dropping items and not returning them to you - great for a normal client, bad for audit. What's actually failing underneath that is `find_device`.

What do you think about the following? A pretty minimal subclass could catch the information you're after:

```python
class AuditClient(Client):
    def find_device(self, *args, **kwargs):
        try:
            return super().find_device(*args, **kwargs)
        except Exception as exception:
            return BadDeviceMarker(exception, args=args, kwargs=kwargs)

client = AuditClient.from_config(...)
result = client.find_device(_id='bad_device')
assert isinstance(result.item, BadDeviceMarker)
```

Still gives back a search result, then you can filter through the list looking for `BadDeviceMarker` or something like that.

An alternative would be something like `Client.from_config(debug_mode=True)` or `client.debug_mode = True` to get those additional items - I'm not very fond of that though as it makes search results potentially inconsistent based on just a simple flag.

cristinasewell:
I took this out

 >I don't think the `--config` is necessarily a requirement as setting `HAPPI_CFG` could be sufficient. Since `Client.from_config()` lets you pass in a configuration file, it seems easy enough to support it here though if it seems useful.

I've been using `client.backend.all_devices` to get all the items  in the database directly (that's probably not the right approach) for some of the validation functions, but for others like testing the `args` and `kwargs` and `validating_device_class` I've been using `Client.all_items`.
Also, when testing with `enforce_value()` for example - I have to use the `happi.item.EntryInfo` to test against so I need to get back a `HappiRegistry` - and if something goes bad here and it can't create a container .... then again I can't test all those items....

The `validate()` will go through all the entries and validate them, and I catch all the errors here, but I know it is not enough...


I will give your suggestion a try - thank you very much :) 
Even with this though, I am still not sure I will be able to test everything (like the `enforce_value()` for example) (but need to give it a try first)

> Hmm, I see. Restating what we're after here: auditing needs to know about both items that can load (to check their information) and items that can't be loaded (to see what the problem is). For the latter, `_get_search_results` is dropping items and not returning them to you - great for a normal client, bad for audit. What's actually failing underneath that is `find_device`.
> 
> What do you think about the following? A pretty minimal subclass could catch the information you're after:
> 
> ```python
> class AuditClient(Client):
>     def find_device(self, *args, **kwargs):
>         try:
>             return super().find_device(*args, **kwargs)
>         except Exception as exception:
>             return BadDeviceMarker(exception, args=args, kwargs=kwargs)
> 
> client = AuditClient.from_config(...)
> result = client.find_device(_id='bad_device')
> assert isinstance(result.item, BadDeviceMarker)
> ```
> 
> Still gives back a search result, then you can filter through the list looking for `BadDeviceMarker` or something like that.
> 
> An alternative would be something like `Client.from_config(debug_mode=True)` or `client.debug_mode = True` to get those additional items - I'm not very fond of that though as it makes search results potentially inconsistent based on just a simple flag.

PR #171, opened 2020-09-21
ENH: bash/fzf-based fuzzy finding of happi items

klauer:
## Description
* Adds `happi search --json` option to output JSON instead of a table.
* Moves IPython import to where it's needed, saving approximately half a second on any other `happi` CLI invocation
* Adds activate/deactivate scripts which can be used in conjunction with a fuzzy finder called `fzf`

Old video from issue #98: https://asciinema.org/a/358799
The video shows the underlying mechanism but not the (relatively) user-friendly option that this PR provides:
with this PR, you type `happi search **[TAB]` and it'll show that same filterable list of names.

## Motivation and Context
* Names are difficult to type and remember, even with globs

## How Has This Been Tested?
* Locally and manually

### DIFF
--- a/conda-recipe/activate.sh
+++ b/conda-recipe/activate.sh
@@ -1,0 +1,17 @@
+# This activates fzf-based fuzzy finding for happi searches.
+# This is meant to be sourced from bash.
+
+_fzf_complete_happi() {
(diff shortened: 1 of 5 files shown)


klauer:
Have to update conda-forge recipe with the new activation scripts

PR #176, opened 2020-10-14
Questionnaire cleanup

### LINKED ISSUE #175: Questionnaire: load failures give no clue as to what happened
## Current Behavior
```
ERROR    Unable to create an object from Questionnaire table motors row 1
```
_Why?_

There's no way to figure out why this didn't load without copy/pasting the questionnaire backend internals (or knowing what it should be set as in the questionnaire website).

## Expected Behavior
```
ERROR    Unable to create an object from Questionnaire table motors row 1 because
           (the field X was not set properly | because class Y could not be loaded |
            because some other reason)
```

## Possible Solution
* Add debugging information to questionnaire loading messages
* Add some traceable elements to the backend so superusers can poke around with the information (`qs.failed_devices` or `qs.stashed_exceptions`, `qs.raw_info`, etc.)

klauer:
## Description
* Add a questionnaire helper class to:
1. Break up loading routine into more maintainable functions
2. Allow for introspection of state

## Motivation and Context
Closes #175 

## How Has This Been Tested?
* Locally, but I'm hoping to add more tests
* Looking for an experiment that has valid entries for better testing, too

```python
In [1]: import happi
from happi.backends.qs_db import QSBackend

qs = QSBackend('xcsx41018', use_kerberos=True)

In [2]: qs.helper
Out[2]: <QuestionnaireHelper experiment=xcsx41018 run_number=run18 proposal=X410 beamline=XCS>
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,12 +2,27 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.5.0
+    rev: v3.2.0
     hooks:
+    -   id: no-commit-to-branch
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
+    -   id: check-ast
+    -   id: check-case-conflict
+    -   id: check-json
+    -   id: check-merge-conflict
+    -   id: check-symlinks
+    -   id: check-xml
+    -   id: check-yaml
+        exclude: '^(conda-recipe/meta.yaml)$'
+    -   id: debug-statements
 
--   repo: https://gitlab.com/pycqa/flake8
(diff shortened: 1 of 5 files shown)


klauer:
The questionnaire may not have been working after the happi refactors earlier in the year.

In any case, where we are:
1. Tests are back to passing, with a bunch of deprecation warnings due to our usage of `happi.containers.Device` (in `pcdsdevices.happi.containers.Motor` et al)
2. Considering a different "translation" dictionary here (see diff); am very open to suggestions
3. The new helper lets you poke around a bit and see what's happening

At minimum for review, please take a look at the log levels and messages and see if you agree so we can close out the issue.

ZLLentz (approved):
This looks clean and well-motivated. No red flags.

PR #180, opened 2020-10-28
Create epics arch

cristinasewell:
Created a new function in the `QuestionnaireHelper` that writes the aliases and pvs to a file.

## Motivation and Context

Trying to close an issue but I might be very off here 🌁 .... is this the right place for this implementation?

## How Has This Been Tested?

Locally only.

N/A

### DIFF
--- a/happi/backends/qs_db.py
+++ b/happi/backends/qs_db.py
@@ -5,6 +5,7 @@ import functools
 import logging
 import re
 from typing import Optional
+import os
 
 from psdm_qs_cli import QuestionnaireClient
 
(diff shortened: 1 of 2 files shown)


klauer:
Let's think about where this might belong...

LCLS-specific things:
* Questionnaire
* hutch-python
* pcdsdevices and other pcds* repos

Repositories other facilities may currently be using or may use in the future:
* happi
* typhos, pydm, etc.

While it may not be incorrect to keep this method in the questionnaire backend, I think it might be more appropriate somewhere in the hutch-python repository, interacting with the happi [_client_](https://github.com/pcdshub/hutch-python/blob/e910102fa840b343e7a004a96827983b60fbe40a/hutch_python/qs_load.py#L66-L68) instead of the backend.
Interacting with the client also ensures you only get valid entries out.

ZLLentz:
After some thought, I agree with Ken. I think this needs to manifest itself as a command-line script that we call on-demand from a pcds-specific repository. `hutch-python` is a good candidate.

cristinasewell:
> After some thought, I agree with Ken. I think this needs to manifest itself as a command-line script that we call on-demand from a pcds-specific repository. `hutch-python` is a good candidate.

Sounds good - thank you guys, I'm going to close this pr then and move to `hutch-python`

PR #191, opened 2020-12-21
[WIP] add YAQItem class

untzag:
## Description

Adds support for [yaq](https://yaq.fyi) to happi.

## Motivation and Context

Would like to use happi to manage yaq devices alongside other bluesky interface sources :heart:.

## How Has This Been Tested?

Tested locally. Would love to add tests but would like some direction from maintainers. We have [yaqd-fakes](https://gitlab.com/yaq/yaq-python/-/tree/master/yaqd-fakes) so it'll be easy to stand up some daemons for testing purposes. Marked as WIP because I'd like to see tests added before merge.

Willing to document, would like some direction from maintainers.

### DIFF
--- a/happi/item.py
+++ b/happi/item.py
@@ -363,3 +363,11 @@ class OphydItem(HappiItem):
     args.default = ['{{prefix}}']
     kwargs = copy.copy(HappiItem.kwargs)
     kwargs.default = {'name': '{{name}}'}
+
+
+class YAQItem(HappiItem):
+    port = EntryInfo("TCP port.", enforce=int, optional=False)
+    host = EntryInfo("Host.", optional=True, default="localhost")
+    kwargs = copy.copy(HappiItem.kwargs)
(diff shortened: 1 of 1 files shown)


klauer:
@untzag - happy to see you're interested in using happi.

For historical reasons we have some LCLS-specific containers inside of happi, but our general recommended approach from here on out is to add a `happi.containers` entrypoint to _your_ package. This is unfortunately not yet documented as far as I recall. Our primary example would be in pcdsdevices:

The entrypoint definition:
https://github.com/pcdshub/pcdsdevices/blob/e5cf6f66c9db5d5d32bb4a654f76a74e17f3e5ab/setup.py#L13-L15

And the sub-package containing those classes:
https://github.com/pcdshub/pcdsdevices/blob/master/pcdsdevices/happi/containers.py

untzag:
@klauer thanks so much these hints are very helpful :smile: 

I love that y'all are using entry points for this, what a good idea

klauer:
No problem, @untzag!

PR #199, opened 2021-01-08
FIX: HappiRegistry call load again if requested item not found

### LINKED ISSUE #198: confusing behavior at import time for packages implementing entry points
## Expected Behavior

The happi registry leads to some very confusing behavior when trying to load HappiItem subclasses using entry points. The surface-level observation is that the registry seems to be import-order dependent. I'm pretty sure I understand exactly why this happens---only question is how to fix it :thinking:.

## Current Behavior

You can actually cause unexpected behavior using your own pcdsdevices package.

This following works as expected:

```python
>>> import happi
>>> from pcdsdevices.happi import containers
>>> happi.containers.registry._registry   # contains pcdsdevices
```

The following does NOT work as expected:

```python
>>> from pcdsdevices.happi import containers
>>> import happi
>>> happi.containers.registry._registry   # does NOT contain pcdsdevices
```

## Possible Solution

Unexpected behavior occurs when happi is _first_ imported by the container-defining module. If this happens, your `HappiRegistry` singleton will instantiate DURING module import, and BEFORE the container classes themselves are defined. Conversely, if happi is imported before Python touches the container-defining module, happi itself will load the module properly within `HappiRegistry.load`, and everything works as expected.

Minimally invasive way to fix this is to have the registry try to load again when an item is not found. I've got a working version of this, PR coming soon.

I will ask---why does the registry work this way? Couldn't we just have items be lazy-loaded on the client level?

## Context

Unlike pcdsdevices, my package yaqc-bluesky intends to define the happi containers as a top-level module that's imported with the top-level package. I guess I don't _have_ to do it this way, but it's my preference. Given this design, I was immediately hit with very confusing import-order dependence between happi and my top-level module. I imagine you may never have to deal with this for pcdsdevices, since the containers are defined within a subpackage.

untzag:
## Description

resolves #198 

## Motivation and Context

see #198 

## How Has This Been Tested?

Tested locally. Willing to add tests but I need guidance from maintainers :sweat_smile: 

Willing to add documentation but I need guidance from maintainers.

### DIFF
--- a/happi/containers.py
+++ b/happi/containers.py
@@ -1,9 +1,10 @@
-import logging
 import inspect
+import logging
+
 import entrypoints
 
-from .item import HappiItem, OphydItem
 from .device import Device
+from .item import HappiItem, OphydItem
 
 logger = logging.getLogger(__name__)
 
@@ -33,15 +34,21 @@ class HappiRegistry:
         return cls.__instance
 
     def __getitem__(self, item):
+        if item not in self._registry:
+            self.load()
         return self._registry.get(item)
 
     def __contains__(self, item):
+        if item not in self._registry:
+            self.load()
         return item in self._registry
 
     def items(self):
         yield from self._registry.items()
 
     def entry_for_class(self, klass):
+        if klass not in self._reverse_registry:
+            self.load()
         return self._reverse_registry.get(klass)
 
     def _safe_add(self, entry_name, klass):


ZLLentz:
I think this is a reasonable fix. We'll document it in the release notes, it should not change any behavior for users outside of fixing this failure case.

I'm trying to think if it's enough, or even if the original approach here had some flaw and needs revisiting based on the issue you posted. Import order shenanigans are the worst.

If it fixes your case it's worth merging/tagging and thinking more about this in some follow-up. I'll merge if Ken or Hugo agrees.

ZLLentz (approved):

klauer (approved):
I added some more thoughts in the original issue. I think this is a fine short-term fix, but our implementation itself needs a good bit more of work.

ZLLentz:
Ok, we are in agreement then. I will merge and push out a bugfix tag.

PR #203, opened 2021-01-24
FIX: expand home directory (~) in JSON backend database path

untzag:
## Description

Pass jsondb path through `os.path.expanduser` .

## Motivation and Context

I expect the following cfg file to work:

```ini
[DEFAULT]
path = ~/.local/share/happi/db.json
```

Currently it does not, happi requires FULL path.

## How Has This Been Tested?

Tested locally.

Not documented.

### DIFF
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -47,7 +47,7 @@ class JSONBackend(_Backend):
     """
 
     def __init__(self, path, initialize=False):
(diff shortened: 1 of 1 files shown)


klauer (approved):
Thanks - one path-related bug down, several more to go. 👍

ZLLentz (approved):
Thank you for the fix @untzag!

PR #207, opened 2021-01-24
happi update <json>

### LINKED ISSUE #190: CLI tools: update, dump
## Expected Behavior

It would be great for me if the happi CLI supported richer ways for external applications to interact with the database. These are intended as an API, not necessarily a tool for humans.

## Current Behavior

N/A

## Possible Solution

I would love to see happi support the following:

```bash
$ happi update <json>
```

Here, applications would feed happi a JSON-formatted version of the database contents. Happi would add these new items to the database. Any items that already exist would be overwritten with the new values. Just like `dict.update` in Python. Happi should verify items are formatted correctly before committing to the database.

```bash
$ happi dump
<json to standard out>
```

Literally just dump the whole database as a JSON string to standard out. The output of dump should be able to transparently pipe into update---in this way it will become easy to concatenate databases.

N/A

## Context

I'm trying to use happi to manage [yaq](https://yaq.fyi) devices. In yaq-land, I already have tooling that allows me to see what's installed. All I need to do is reformat this information and dump it into happi. Since yaq is not necessarily Python, I'd prefer to do this through an entry point rather than the Python API.

## Your Environment

N/A

untzag:
## Description

Add new CLI sytntax: `happi update <json>`. User supplies a JSON payload defining Happi items. Items from payload are verified then splatted into the database. In the case of an `_id` conflict, the NEW item is preferred and the old item is overwritten. Name is inspired by Python's `dict.update` method.

Closes #190.

## Motivation and Context

This syntax allows for bulk management of the Happi database through easy integration with other command line tools. I am simultaneously updating yaqd to introduce Happi-formatted output, see [GitLab MR](https://gitlab.com/yaq/yaqd-control/-/merge_requests/42).

It is my hope that this interface will allow for other device-discovery tools to easily add Happi support.

## How Has This Been Tested?

Tested locally. All of the following work on my machine.

```bash
$ yaqd list -f happi | happi update
$ happi search '*' --json | happi update
$ happi update "$(yaqd list -f happi)"
$ happi update "$(happi search '*' --json)"
```

I want to write tests for Happi CI! I'd like some guidance from maintainers if possible.

Willing to document, would like guidance from maintainers :sweat_smile: 

In addition to documenting `happi update` itself, I'd like to encourage documentation of the "standardized" HappiItem JSON representation. This will allow other developers to confidently create tooling for this interface.

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -58,6 +58,11 @@ def get_parser():
                                         'given device loaded.')
     parser_load.add_argument('device_names', nargs='+',
                              help='The names of one or more devices to load')
+    parser_update = subparsers.add_parser("update",
+                                          help="Update happi db "
+                                          "with JSON payload.")
+    parser_update.add_argument("json", help="JSON payload.",
+                               default="-", nargs="*")
     return parser
 
 
@@ -252,6 +257,19 @@ def happi_cli(args):
 
         from IPython import start_ipython  # noqa
         start_ipython(argv=['--quick'], user_ns=devices)
+    elif args.cmd == "update":
+        # parse input
+        input_ = " ".join(args.json).strip()
+        if input_ == "-":
+            items_input = json.load(sys.stdin)
+        else:
+            items_input = json.loads(input_)
+        # insert
+        registry = happi.containers.registry
+        for item in items_input:
+            item = client.create_device(device_cls=item["type"], **item)
+            exists = item["_id"] in [c["_id"] for c in client.all_items]
+            client._store(item, insert=not exists)
 
 
 def main():
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -265,7 +265,6 @@ def happi_cli(args):
         else:
             items_input = json.loads(input_)
         # insert
-        registry = happi.containers.registry
         for item in items_input:
             item = client.create_device(device_cls=item["type"], **item)
             exists = item["_id"] in [c["_id"] for c in client.all_items]
--- a/happi/tests/test_cli.py
+++ b/happi/tests/test_cli.py
@@ -321,3 +321,35 @@ def test_load(caplog, happi_cfg):
         m.assert_called_once_with(argv=['--quick'], user_ns=devices)
     with caplog.at_level(logging.INFO):
         assert "Creating shell with devices" in caplog.text
+
+
+def test_update(happi_cfg):
+    new = """[ {
+        "_id": "TST_BASE_PIM2",
+        "active": true,
+        "args": [
+            "{{prefix}}"
+        ],
+        "beamline": "TST",
+        "creation": "Wed Jan 30 09:46:00 2019",
+        "device_class": "types.SimpleNamespace",
+        "kwargs": {
+            "name": "{{name}}"
+        },
+        "last_edit": "Fri Apr 13 14:40:08 2018",
+        "macros": null,
+        "name": "tst_base_pim2",
+        "parent": null,
+        "prefix": "TST:BASE:PIM2",
+        "screen": null,
+        "stand": "BAS",
+        "system": "diagnostic",
+        "type": "OphydItem",
+        "z": 6.0,
+        "from_update": true
+    } ] """
+    happi.cli.happi_cli(["--verbose", "--path", happi_cfg, "update", new])
+    #
+    client = happi.client.Client.from_config(cfg=happi_cfg)
+    item = client.find_device(name="tst_base_pim2")
+    assert item["from_update"]


ZLLentz:
Sorry for the late review here- I saw this come in, but I didn't see when you edited the WIP out of the title.

ZLLentz (approved):
The idea and implementation here are both solid.

The way you've gone about this shows a few holes in the client API that I'd like to resolve at some point... there should never be a reason to call `_store` from a different module, but here we are, that's the correct way to do it.

ZLLentz:
@klauer I'd like you to give this a peek before I merge, you've been in the internals here more than I have

untzag:
I'd love to add tests and docs before merge. I see you have a pytest fixture... how should we test this functionality?

untzag:
Also I'm wondering if there should be a flag `--no-overwrite` for this feature. Or perhaps the default should be not to overwrite, with `--overwrite` being the flag...

untzag:
>The way you've gone about this shows a few holes in the client API that I'd like to resolve at some point... there should never be a reason to call _store from a different module, but here we are, that's the correct way to do it.

If I was y'all I would add a kwarg `exist_ok` to `client.add_device` for this. Default `False`.

happi/cli.py
```diff
@@ -252,6 +257,19 @@ def happi_cli(args):
...
         start_ipython(argv=['--quick'], user_ns=devices)
+    elif args.cmd == "update":
+        # parse input
+        input_ = " ".join(args.json).strip()
+        if input_ == "-":
+            items_input = json.load(sys.stdin)
+        else:
+            items_input = json.loads(input_)
+        # insert
+        registry = happi.containers.registry
```
klauer, line 268: `registry` unused here
```suggestion
```

happi/cli.py
```diff
@@ -252,6 +257,19 @@ def happi_cli(args):
 
         from IPython import start_ipython  # noqa
         start_ipython(argv=['--quick'], user_ns=devices)
+    elif args.cmd == "update":
+        # parse input
+        input_ = " ".join(args.json).strip()
+        if input_ == "-":
+            items_input = json.load(sys.stdin)
+        else:
+            items_input = json.loads(input_)
```
klauer, line 266: Just to recap for `update` JSON variable args:
1. `-` to get input from stdin, a great feature
2. Assumes JSON string with optional quoting (as it effectively does something like`json.loads(' '.join(sys.argv[2:])`)

CLI API-wise I personally prefer requiring either pipe or filename (making supporting this really simple with a single https://docs.python.org/3/library/argparse.html#argparse.FileType - which supports files and stdin by way of `-`). 

Not a dealbreaker for merging this PR, and only reflects my opinion.

happi/cli.py
```diff
@@ -252,6 +257,19 @@ def happi_cli(args):
...
+        # insert
+        registry = happi.containers.registry
+        for item in items_input:
+            item = client.create_device(device_cls=item["type"], **item)
+            exists = item["_id"] in [c["_id"] for c in client.all_items]
+            client._store(item, insert=not exists)
```
klauer, line 272: happi has some weird saving methodology, where it patches on `save` to the ~device~ item:
https://pcdshub.github.io/happi/master/client.html#editing-device-information

I think the original API intends for this to be the only call you need:
```suggestion
            item.save()
```

I suppose this may fail with the scenario `insert=True/exists=True` and as such the proposed `exists_ok` kwarg?

klauer (commented):
For testing, following the current pattern of `happi.cli.happi_cli()` with some parametrized arguments should be good enough. https://github.com/pcdshub/happi/blob/master/happi/tests/test_cli.py

Any documentation would be appreciated - we still have a lot of work to get it up-to-speed ourselves.

(also, I didn't see the title change and only noticed this with @ZLLentz's comments)

ZLLentz:
> If I was y'all I would add a kwarg exist_ok to client.add_device for this. Default False.

This seems reasonable. The other thing I was considering was just adding a `client.update(dict)` method.

happi/cli.py
untzag: @klauer I tried this, but I couldn't get it to work the way I want to

1) default behavior to be reading from stdin
2) able to use "$()" syntax

on the other hand you can always use files with the current implementation... `cat myfile.json | happi update` or `happi update "$(cat myfile.json)"`

happi/cli.py
untzag: Correct. This was the least invasive solution I could find to get my desired behavior of overwrite if exists. Your suggestion does work for new items.

untzag:
I see that you auto-build documentation for the cli. This is alright for now in my opinion.

untzag:
@ZLLentz @klauer I believe I'm ready for merge, please take a look when you have a chance

ZLLentz:
I still think this is :+1:. I'm going to make an issue for expanding the client API and if Ken agrees that you've addressed his comments then we should merge.

happi/cli.py
klauer: Right. My suggestion explicitly would not allow that additional input method, as I believe it should be done through a pipe (or a file).

@ZLLentz it's worth noting that if we are to move forward with this specific method, we will have to duplicate this functionality in all other tools we intend to support JSON with.

klauer:
Good enough for me. If you're happy, click the button @ZLLentz.

PR #211, opened 2021-02-10
ENH: allow short and long names

### LINKED ISSUE #206: HappiItem rejects names under three charachters
## Expected Behavior

I commonly call my devices simple things like `d1` or `wa`. I was surprised to find that the HappiItem baseclass rejects these names. Your regex for name, `[a-z][a-z\_0-9]{2,78}$`, requires at least 3 characters.

## Current Behavior

Is enforcing at least 3 characters necessary? I would have expected that any valid Python identifier would be a valid HappiItem name.

Since we're talking about it, you're also enforcing lower case... is that necessary?

## Possible Solution

 `[a-z][a-z\_0-9]{0,78}$`

I also understand that I can change this behavior within my `YAQItem` subclass, and am happy to do that instead if maintainers wish to keep these restraints for the base class.

ZLLentz:
## Description

- Remove the requirement for names to be between 3 and 80 characters
- Keep requirement for names to be valid Python identifiers
- Add previously implicit requirement for names to not conflict with reserved keywords
- Add feature to use arbitrary functions as enforce parameters

## Motivation and Context

This is a site-specific assumption that should not be baked in to the base class
closes #206 

## How Has This Been Tested?

~Online regex tester site to quickly vet python identifier~
The above is no longer valid, since we are no longer using regex here.
But I think this is fine since the existing tests still pass and we're using core python built-ins to validate the names, rather than custom regex.

Added some to the docs folder
Will be in the release notes

### DIFF
--- a/docs/source/containers.rst
+++ b/docs/source/containers.rst
@@ -96,14 +96,15 @@ real value is given it will be checked against the specified ``enforce``
 keyword, raising ``ValueError`` if invalid. Here is a table for how the
 :class:`.EntryInfo` check the type
 
-=======   ===========================
-Enforce   Method of enforcement
-=======   ===========================
-None      Any value will work
-type      type(value)
-list      list.index(value)
-regex     regex.match(value) != None
-=======   ===========================
+========   ===========================
+Enforce    Method of enforcement
+========   ===========================
+None       Any value will work
+type       type(value)
+list       list.index(value)
+regex      regex.match(value) != None
+function   function(value)
+========   ===========================
 
 Fields that are important to the item can be marked as mandatory with
 ``optional=False`` and should have no default value.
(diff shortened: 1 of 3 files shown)


klauer (approved):
I wonder if `enforce=str.isidentifier` could be used here?

ZLLentz:
Would require a small change to the way happi entries do their enforcement, it doesn't let you pass arbitrary functions at the moment

ZLLentz:
After that change, we would probably just want this to be the built-in identifier check you've cited followed by a check for not-a-python-keyword

klauer:
You're right. I really thought it allowed for callables, and am very surprised... And good call on the 'not-a-keyword' as well.

ZLLentz:
I think I just implement this now, we'll want it at some point for other things too.

klauer (approved):
Awesome - a small change on the surface, but it makes things so much more flexible.

ZLLentz:
Fixed the style issue- will proceed to merge

PR #212, opened 2021-02-10
MNT: fix confusing behavior with boolean enforcement

### LINKED ISSUE #187: happi edit command not working for boolean variables
## Expected Behavior
Edit command should edit the field properly

## Current Behavior
This seems to be broken for booleans

Run `happi edit xpp_pol_wp lightpath=False`
xpp_pol_wp entry has `lightpath=True`

## Context
Trying to add laser motor to xpp's beamline

## Your Environment
pcds-3.4.1 on xcs-control

ZLLentz:
## Description

Add a special case for boolean entry enforcement where we interpret the input string instead of just casting it into a boolean.
Accepts strings like "true" and "false" as stand-ins for `True` and `False` in situations where the input must be given as a string, such as in the cli.

## Motivation and Context

`happi edit` and `happi add` do not set boolean values as one would expect
closes #187

## How Has This Been Tested?

Added tests

Will include in release notes, I'm going to count this as a bugfix.

### DIFF
--- a/happi/item.py
+++ b/happi/item.py
@@ -98,8 +98,8 @@ class EntryInfo:
         elif self.enforce is bool and isinstance(value, str):
             # Special case for booleans, since the value may come in as a
             # string, and bool('False') evaluates to True
-            true_values = ['true', 'yes']
-            false_values = ['false', 'no']
+            true_values = ['true', 't', 'yes', 'y']
+            false_values = ['false', 'f', 'no', 'n']
             if value.lower() in true_values:
                 return True
             elif value.lower() in false_values:
--- a/happi/item.py
+++ b/happi/item.py
@@ -95,6 +95,19 @@ class EntryInfo:
         if not self.enforce or value is None:
             return value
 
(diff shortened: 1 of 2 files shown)


happi/item.py
```diff
@@ -95,6 +95,19 @@ def enforce_value(self, value):
         if not self.enforce or value is None:
             return value
 
+        elif self.enforce is bool and isinstance(value, str):
+            # Special case for booleans, since the value may come in as a
+            # string, and bool('False') evaluates to True
+            true_values = ['true', 'yes']
```
klauer, line 101: Can I be lazier?
```suggestion
            true_values = ['true', 'yes', 'y']
```

klauer (approved):
Good fix 👍  `y/n` could be a minor tweak if you agree.

happi/item.py
ZLLentz: lol

ZLLentz: I'll add t, f, y, n

PR #215, opened 2021-02-26
MNT: add handling of mpod questionnaire

cristinasewell:
## Description
Add support to create mpod channels from the questionnaire entries.

## Motivation and Context

This change goes with the additions from this PR:  https://github.com/pcdshub/pcdsdevices/pull/690

## How Has This Been Tested?

Only locally.

N/A

### DIFF
--- a/happi/tests/test_backends.py
+++ b/happi/tests/test_backends.py
@@ -1,10 +1,10 @@
 import fcntl
 import os
 import os.path
+from unittest.mock import patch
 
 import pytest
 import simplejson
-
 from happi import Client
 from happi.backends.json_db import JSONBackend
 from happi.errors import DuplicateError, SearchError
@@ -163,7 +163,7 @@ def test_json_initialize():
 
 @requires_questionnaire
 def test_qs_find(mockqsbackend):
-    assert len(list(mockqsbackend.find(dict(beamline='TST')))) == 14
+    assert len(list(mockqsbackend.find(dict(beamline='TST')))) == 16
     assert len(list(mockqsbackend.find(dict(name='sam_r')))) == 1
 
 
@@ -171,7 +171,7 @@ def test_qs_find(mockqsbackend):
 @requires_pcdsdevices
 def test_qsbackend_with_client(mockqsbackend):
     c = Client(database=mockqsbackend)
-    assert len(c.all_items) == 14
+    assert len(c.all_items) == 16
     assert all(
         d.__class__.__name__ in {'Trigger', 'Motor', 'Acromag', 'LCLSItem'}
         for d in c.all_items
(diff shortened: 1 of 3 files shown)


happi/tests/test_backends.py
```diff
@@ -203,3 +204,23 @@ def test_beckoff_axis_device_class(mockqsbackend):
...
+    # patch the get_max_voltage to be able to determine if LV or HV
+    # if voltage < 50 => LV Channel
+    with patch('pcdsdevices.mpod.MPODChannel.get_max_voltage',
+               return_value=49):
+        d = load_devices(*c.all_devices, pprint=False).__dict__
+        led = d.get('led')
+        assert led.__class__.__name__ == 'MPODChannelLV'
+
+    # if voltage > 50 => HV Channel
+    # with patch('pcdsdevices.mpod.MPODChannel.get_max_voltage',
```
cristinasewell, line 222: Anybody knows how to 'cleanup' or 'close' in a way a mock when using `with` patch?
It seems like the first patch is leaking over to the second one, even if I make a new test function.

I can avoid using `with` and use the `monkeypatch` or the `mocker.patch.object` instead and I think it would work fine, but I did not want to have to import `pcdsdevices` module in this file 😀

klauer: I like to use a pattern similar to:

```python
@pytest.mark.fixture(scope='function')
def my_modified_obj(monkeypatch):
   obj = ...
   monkeypatch.setattr(obj, 'attr', 'value')
   yield obj
   # any cleanup if necessary, the `yield` above makes this fixture a context manager effectively

def test_mytest(my_modified_obj):
    assert my_modified_obj.attr == 'value'
```

It keeps the monkeypatch scoped to a specific test and is pretty explicit about it.

That said, the `with` block you use should have reset the original attribute value - not sure why it didn't, or if maybe something else is going on there.

cristinasewell: Thank you Ken, yeah I think something else is going on, only if I test them separately it works though.
I am not calling this function that I need to mock (`get_max_voltage`) anywhere directly here  - I probably have to mock something else instead to be able to test both option

cristinasewell: Coming back to this guy, I tested it with some dummy PVs and no mocking `get_max_voltage` and it seems like it is working, which is good:
```
'led': MPODChannelLV(csewell:mpod:l:CH:001, name=led), 'blower': MPODChannelHV(csewell:mpod:h:CH:001, name=blower)}
MPODChannelLV(csewell:mpod:l:CH:001, name=led)
MPODChannelHV(csewell:mpod:h:CH:001, name=blower)
```

Later I realized that I had to use `use_cash=False` in order for it to work fine:
```
d = load_devices(*c.all_devices, pprint=False, use_cache=False).__dict__
```

I was getting here:
https://github.com/cristinasewell/happi/blob/d5c24dc8c4416d852f98d4e1c0b3dc62d9a4c6a5/happi/loader.py#L122

 if I did not set `use_cash=False` , and the second mpod was created the same as the first one (E.g.: if the first one was HV the second one was HV too ... 

I am not sure if I understand why was this happening exactly ... it is probably a combination of how the `mockqsbackend` is setup and the fact that I am looking at the voltage to create the mpod channel ....

Issue #231, opened 2021-12-22
heuristic loading


untzag:
## Expected Behavior

Would be nice for me if `happi load "*"` worked, in analogy to `happi search "*"`.

## Current Behavior

`happi load` requires users to explicitly specify which hardware to load. None of the search functionality is supported.

## Possible Solution

Would be nice to provide support for searching and loading in one user-friendly expression.

untzag:
The obvious idea is just to route the arguments to `load` through the same code path as `search`.

Problem is that `happi search <criteria>` combines multiple criteria with AND. So current syntax of `happi load name1 name2 name3` would result in nothing loaded because no items have all three of those names. `load` needs OR logic to work as a user expects, but then it becomes impossible to narrow down to a specific item based on multiple criteria. Maybe the OR logic would only apply to names?

Perhaps #232 is the best answer after all :thinking:. My only problem is that most users don't really understand the shell enough to use tricks like that (I myself mess up all the time...). Simply dropping into a repl with your objects is something I want everyone to be comfortable with.

@klauer @ZLLentz tell me what to do please. I'll implement it and make a PR, I'm just not sure what the right approach is.

ZLLentz:
You're right, the OR vs AND conflict here makes this tricky, and most solutions would probably make the CLI confusing (which is a problem, since the point of this is to make the invocation more user-friendly).

I'd like to hear Ken's opinion on this when he's back from his vacation. My initial thought would be to add some syntax like:
```
happi load --search "*"
```
And we'd be able to compose multiple of them like:
```
happi load name1 name2 --search name="*_mot" --search class="*Motor"
```
Where the above would load in the devices named `name1` and `name2`, all devices whose names end in `_mot`, and all devices whose classes end in `Motor` (and only one of each device if they show up in multiple searches)

But looking at the options in argparse this seems awkward for more complicated searches...

klauer:
To get a truly flexible query, we'd need some sort of `x AND (y OR NOT z)`-style grammar which would be rather excessive (in my opinion) for a tool like happi. I have yet to see a nice implementation of this sort of thing in a CLI tool - see `find` for a quick example, with its chainable expressions by way of `-and` and `-or`.

@ZLLentz 's suggestion above seems fine, though the implicit "OR" between the names/`--search` items may not be desirable. 

I think the real question here is what will the most common use cases here be? And how can we cover that with minimal added complexity?

[...]

PR #232, opened 2021-12-22
ENH: cli search flag to output names

untzag:
## Description

Enable new flag to `happi search`, `--names`.

```bash
$ happi search "*"
(pretty table)
& happi search "*" --names
d1 d2 mysensor
```

If both `--json` and `--names` are provided the output is json.

Feel free to reject or ask for major changes.

## Motivation and Context

Motivated by #231, I can now do this:

```bash
$ happi load "$(happi search "*" --names)"
[2021-12-22 12:04:39] - INFO -  Creating shell with devices ['d1 d2 mysensor']
Python 3.9.9 | packaged by conda-forge | (main, Dec 20 2021, 02:41:03) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.30.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: d1
Out[1]: <yaqc_bluesky._device.YAQDevice at 0x7f41ed902f10>
```

I had to massage `load` a little bit to do this, but not in a way that changes existing behavior.

I don't think this closes that PR, as I still would prefer direct syntax if possible.

## How Has This Been Tested?

Works on my machine :wink:

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -38,6 +38,8 @@ def get_parser():
                                           'database.')
     parser_search.add_argument('--json', action='store_true',
                                help='Show results in JSON format.')
+    parser_search.add_argument('--names', action='store_true',
+                               help='Return results as whitespace-separated names.')
     parser_search.add_argument('search_criteria', nargs='+',
                                help='Search criteria of the form: '
                                'field=value. If "field=" is omitted, it will '
@@ -168,6 +170,9 @@ def happi_cli(args):
         if args.json:
             json.dump([dict(res.item) for res in final_results], indent=2,
                       fp=sys.stdout)
+        elif args.names:
+            out = " ".join([res.item.name for res in final_results])
+            print(out)
         else:
             for res in final_results:
                 res.item.show_info()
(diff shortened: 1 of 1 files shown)


happi/cli.py
```diff
@@ -255,7 +261,9 @@ def happi_cli(args):
         logger.debug('Starting load block')
         logger.info(f'Creating shell with devices {args.device_names}')
         devices = {}
-        for name in args.device_names:
+        names = " ".join(args.device_names)
+        names = names.split()
```
ZLLentz, line 265: I'm happy to merge this (just got back from vacation), quick question: is this so that the whitespace-delimited set of names isn't interpreted as a single string in your load example?

untzag: @ZLLentz thanks for your review. You are exactly correct.

To be honest I think I was being a bit dumb when I made the PR. I have no need for the extra quotes in my command. `happi load $(happi search "*" --names)` works just great without the changes you highlighted.

On the other hand it might be nice default behavior for any strings given as arguments to be whitespace separated before interpretation as item names.

I'm happy keeping or removing the changes in lines 264 265. Just let me know what you prefer.

ZLLentz: I think, even though it's a bit confusing and not strictly necessary for the stated purpose of the PR, it's a good improvement. We force the names to be valid Python identifiers in `HappiItem`, which means we can't have any names with whitespace in them anyway. Others are bound to run into the same thing.

I'll merge this as-is, thank you for the contribution!

ZLLentz (approved):

PR #236, opened 2022-03-08
ENH: optional per-device load times

klauer:
## Description
* Only affects those using `load_device` or `load_devices`, which are CLI-facing wrappers around `load_container` (used by `client[device].get()`
   * Are there any users outside of SLAC for this?
* Cleans and annotates `happi.loader`
* Some warning messages were made more descriptive
* Consolidated message printing logic
* Bonus: updates old pre-commit settings

## Motivation and Context
* Some devices take a long time to load, it would be nice to know which without digging too much

## How Has This Been Tested?
* hutch-python loading and test suite

This PR text

image

### DIFF
--- a/happi/loader.py
+++ b/happi/loader.py
@@ -35,7 +35,7 @@ def fill_template(
     template : str
         Jinja2 template source.
 
-    device : HappItem
+    device : HappiItem
         Any happi item container.
 
     enforce_type : bool, optional
--- a/happi/loader.py
+++ b/happi/loader.py
@@ -5,6 +5,7 @@ import asyncio
 import importlib
 import logging
 import sys
+import time
 import types
 from functools import partial
 from multiprocessing.pool import ThreadPool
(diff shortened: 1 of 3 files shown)


happi/loader.py
```diff
@@ -19,57 +22,80 @@
...
 
     Parameters
     ----------
     template : str
-        Jinja2 template
+        Jinja2 template source.
 
-    device : happi.Device
-        Any device container
+    device : HappItem
```
tangkong, line 38: HappItem > Happ**i**Item ?

I almost like the mispelled version more

tangkong (approved):
Looks good to me.  Nice addition and should help immensely with error reporting

happi/loader.py
ZLLentz: I unironically think happ is a better name then happi, sounds like some new hip database. Have you stored all your devices in happ?

klauer: I'm digging it!

... I wish my typos were always so entertaining and thought-provoking 😁

ZLLentz (approved):
Looks good to me!
The titular feature here is straightforward, I spent some time reviewing the other cleanups more closely and they look good too

PR #237, opened 2022-03-11
MNT: switch fcntl warning to debug

### LINKED ISSUE #220: Remove fcntl warning
The following message on Windows is annoying:

https://github.com/pcdshub/happi/blob/d5cba946cd4a58a0b6a9da011e256f84d9ced81a/happi/backends/json_db.py#L22

This is something that should be part of the documentation, not a message on every invocation of `happi`.

klauer:
## Description
* Reduce on-import warning of fcntl to debug.
* This is shown repeatedly on Windows for ads-deploy - which doesn't even really use happi - and is a bit annoying.

## Motivation and Context
Closes #220 

## How Has This Been Tested?
If it passes flake8 and imports, this should be fine.

Issue / PR text

### DIFF
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -19,7 +19,7 @@ logger = logging.getLogger(__name__)
 try:
     import fcntl
 except ImportError:
-    logger.warning("Unable to import 'fcntl'. Will be unable to lock files")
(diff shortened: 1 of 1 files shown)


ZLLentz (commented):
LGTM (late)

I actually see this repeatedly in my normal workflow

Issue #202, opened 2021-01-19
Some dependencies in conda recipe are optional


klauer:
## Current Behavior
Some dependencies are optional (such as `psdm_qs_cli`, `pymongo`), but we require them in our conda recipe.

## Expected Behavior
Optional dependencies should not turn into requirements in our conda recipe.

## Possible Solution
Utilize `run_constrained` to constrain versions of these optional dependencies.
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained

## Context
Recipe: https://github.com/pcdshub/happi/blob/master/conda-recipe/meta.yaml

ZLLentz:
I've never seen this, very interesting. Would we want to set these up as such on the conda-forge recipe too?

klauer:
Yes, especially on the conda-forge recipe - as it's where people would install happi from primarily. 

I'm thinking when we make this change we can also better leverage conda recipe's support for test dependencies and test scripts - further reworking our CI structure. (as we briefly chatted about on Slack)

PR #240, opened 2022-05-10
BLD: move psdm_qs_cli, pymongo to run-constrained reqs

tangkong:
## Description
Moves psdm_qs_cli, pymongo to the run-constrained section of the conda recipe 

## Motivation and Context
#202 

## How Has This Been Tested?
I'm actually not sure how to test this

This PR

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -22,9 +22,10 @@ requirements:
       - entrypoints
       - jinja2
       - prettytable
+      - simplejson
+    run-constrained:
       - psdm_qs_cli
       - pymongo
-      - simplejson
 
 test:
     imports:


klauer (approved):
I think this is probably fine. `run-constrained` without version information seems weird, but it's probably the only way to indicate an optional dependency?

The changed recipe is the one that's used for CI testing and not conda-forge builds.

tangkong:
Actually after doing some reading, it seems conda doesn't really support optional dependencies in the same way pip does.  

There have been quite a few proposals for ways to do something like `conda install happi[dev]`, but none seem to have been taken.  

Looking more closely at run-constrained, it does seem like maybe we should just leave them out at the moment.  At the moment it's somewhat of a noop, essentially saying: "if this package is installed constrain it to be any version"

klauer:
On second thought, I suppose you're right. If there's no canonical way to show optional deps, then we shouldn't misuse this mechanism.

It may be worth considering that we have a minimum compatible version for those dependencies. I don't know what version those are, but we might want to just say "we need (whatever is newest as of today) from now on"

klauer:
pip-based tests are passing now as krtc/psdm_qs_cli are on PyPI as of a couple days ago: https://app.travis-ci.com/github/pcdshub/happi/jobs/569776861

Can we remove it from xfail here?

conda-recipe/meta.yaml
```diff
@@ -22,9 +22,10 @@ requirements:
       - entrypoints
       - jinja2
       - prettytable
-      - psdm_qs_cli
-      - pymongo
       - simplejson
+    run-constrained:
+      - psdm_qs_cli >=0.3.1
+      - pymongo >=4.0.2
```
ZLLentz, line 28: I think something like `extras` configuration is more appropriate for optional dependencies- I don't remember exactly how to do this (not a conda thing).

Requiring a recent API for these makes it easier on us as we move forward, so it's probably correct to do this :+1:. The exact version here doesn't matter too much. We can update these exact numbers as appropriate if we use newer features of these dependencies.

PR #243, opened 2022-05-11
TST: use tempfile for mockjsonclient fixture

### LINKED ISSUE #216: Happi test suite creates json file and fails to clean up
## Expected Behavior

Either clean up the json file or don't create it in the first place

## Current Behavior

Creates a file in the current directory
See https://github.com/pcdshub/happi/blob/d5c24dc8c4416d852f98d4e1c0b3dc62d9a4c6a5/happi/tests/conftest.py#L96

## Context

I'm looking at reworking some of our full environment testing frameworks and noticed this file popping up in my current working directory when running `pytest --pyargs happi`

tangkong:
## Description
- uses tempfile.NamedTemporaryFile instead of writing a .json file and cleaning it up (in conftest.py and test_backends.py)

## Motivation and Context
closes #216 

There are still portions where json is created, but the writing of those files seems to be the target of the test, so I've left those alone

## How Has This Been Tested?
Tests run and no json is left behind.  

This PR

### DIFF
--- a/happi/tests/conftest.py
+++ b/happi/tests/conftest.py
@@ -1,4 +1,5 @@
 import logging
+import tempfile
 from unittest.mock import patch
 
 import pytest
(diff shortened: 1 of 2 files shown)


klauer (approved):
This looks good to me.
I'm a big fan of context manager fixtures that clean up after themselves so nicely. 👍

ZLLentz (approved):
LGTM

Issue #221, opened 2021-08-05
Happi GUI Front-end tools (Qt and/or web-based)


ZLLentz:
## Expected Behavior

Develop a GUI that lets you browse and edit the happi database

klauer:
Web-based or PyQt-based are our top options.

ZLLentz:
Web-based:
- happi server process (with some web auth?)
- web front-end that communicates with the process

PyQt-based:
- local application that runs qt and happi
- auth is based on being able to log in to the system

klauer:
Some browsing support is available already:
https://github.com/pcdshub/happi/blob/master/happi/qt/model.py

atef improves upon this with a happi item + device/component browser. Some of this should be upstreamed to happi:

![image](https://user-images.githubusercontent.com/5139267/167188494-bf2815c4-f1d4-4205-9baf-382bb7c940ad.png)

tangkong:
Should we close this and make more specific issues?  top of mind thanks to #246 
- a web gui might not be absurd once we properly switch to mongo
- editable via gui might be in striking distance after merging in atef changes

klauer:
Sounds good to me 👍

PR #246, opened 2022-05-12
ENH: add happi search widget from atef

klauer:
## Description
Add more user-friendly happi search widget from atef.

## Motivation and Context
We made these for atef, we may as well upstream them to happi.
We need to find a home for the ophyd device tool as well - likely not happi.

image
image


klauer:
(Rebasing on happi master; less because I want to work on this PR and more because I want to see the effect of the line-profiler 4.1.1 release)

klauer:
We might have some updates from atef on these widgets. Perhaps we merge and then follow up?

tangkong:
It does look like there have been some changes.  I'm happi to merge those changes in later.  

This addition does make me wonder if it's worth the time to figure out how to separate happi[ui] from happi[core].  There's not too much to this GUI currently, but I kinda liked the pure-code-ness of happi.  

also if I have to add qt tests to happi I'm gonna cry

klauer:
I suppose we could have `happi-qt` instead? I'm fine with that I guess, but having it get out of sync would be a bit painful

tangkong:
I do think having the viewer in this package will be helpful.  Maybe we reconsider `happi-qt` or `happi[qt]` when we accrue more substantial ui elements.  For now I'm ok with this

ZLLentz:
Should we merge this before or after the upcoming tag?

klauer:
This may be ready to merge? @ZLLentz @tangkong

tangkong (approved):
I'm ok with merging this, with the understanding that we'll have to merge in the updated version from atef and/or create happi-qt later on.

tangkong:
closes #221

PR #249, opened 2022-05-23
ENH: Refactor happi cli to use click instead of argparse

tangkong:
## Description
A rather long PR, which initially started as an effort to simplify our cli code base using `click`, but started to tackle broken/missing functionality along the way.  An effort has been made to preserve existing functionality, except where existing functionality was broken.  

### Enhancements:
- Refactors happi_cli into a click command group, with subcommands (search, add, etc)
- Reworks tests to use click.testing.CliRunner, which keeps track of exit code, exception, and stdout/stderr
- Adds various helpers along the way
- Enforce values received in `happi edit`, and properly handle dicts/lists

### Bug fixes:
- Fixes user-input dictionaries and lists to actually work (issue here?)

## Motivation and Context
The CLI interface has needed some touching up, particularly in the user-prompting side of things.  Users often edit the json directly, and while this may be more direct, it's more prone to error.  If happi provides a good interface for this, perhaps we can preempt some of these errors

Fun notes:
- `click.prompt` has built in type checking and conversion, which lines up very well with `EntryInfo.enforce`.  
- `click.testing.CliRunner` keeps track of exit code, exceptions, and output.  The output capturing actually happens before capsys, but caplog still works.  
- subcommands get passed a `Context` from the main invoking group (in this case `happi_cli`).  To run commands individually, you must create this context manually.  This is what is done in some of the search tests.  In most cases, you can just invoke the main command.  

## How Has This Been Tested?
Test suite has been reworked.  Interactive testing along the way

Docstrings and this PR

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -22,11 +22,13 @@ from .utils import is_a_range, is_valid_identifier_not_keyword
 logger = logging.getLogger(__name__)
 
 version_msg = f'Happi: Version {happi.__version__} from {happi.__file__}'
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
 
 @click.group(
     help=('commands available: search, add, edit, load, update, '
-          'container-registry, transfer')
+          'container-registry, transfer'),
+    context_settings=CONTEXT_SETTINGS
 )
 @click.option('--path', type=click.Path(exists=True),
               help='Provide the path to happi configuration file.')
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -1,93 +1,37 @@
 """
 This module defines the ``happi`` command line interface.
 """
-import argparse
 import fnmatch
 import json
 import logging
 import os
 import sys
 
+import click
 import coloredlogs
 import prettytable
 
 import happi
 
 from .prompt import transfer_container
-from .utils import is_a_range
+from .utils import OptionalDefault, is_a_range
 
 logger = logging.getLogger(__name__)
 
-
-def get_parser():
-    """Defines HAPPI shell commands."""
-    # Argument Parser Setup
-    parser = argparse.ArgumentParser(description='Happi command line tool')
-
-    # Optional args general to all happi operations
-    parser.add_argument('--path', type=str,
-                        help='Provide the path to happi configuration file.')
-    parser.add_argument('--verbose', '-v', action='store_true',
-                        help='Show the debug logging stream.')
-    parser.add_argument('--version', '-V', action='store_true',
-                        help='Show the current version and location of Happi '
-                             'installation.')
-    # Subparser to trigger search arguments
-    subparsers = parser.add_subparsers(help='Subcommands used to search, add, '
-                                       'edit, or load entries', dest='cmd')
-    parser_search = subparsers.add_parser('search', help='Search the happi '
-                                          'database.')
-    parser_search.add_argument('--json', action='store_true',
-                               help='Show results in JSON format.')
-    parser_search.add_argument('--names', action='store_true',
-                               help='Return results as '
-                               'whitespace-separated names.')
-    parser_search.add_argument('search_criteria', nargs='+',
-                               help='Search criteria of the form: '
-                               'field=value. If "field=" is omitted, it will '
-                               'be assumed to be "name". You may include as '
-                               'many search criteria as you like; these will '
-                               'be combined swith ANDs.')
-    parser_add = subparsers.add_parser('add',
-                                       help='Add new entries interactively.')
-    parser_add.add_argument('--clone', default='',
-                            help='Copy the fields from an existing container. '
-                                 'Provide the name of the item to clone.')
-    parser_edit = subparsers.add_parser('edit',
-                                        help='Change an existing entry.')
-    parser_edit.add_argument('name', help='Name of the item to edit')
-    parser_edit.add_argument('edits', nargs='+',
-                             help='Edits of the form: field=value')
-    parser_load = subparsers.add_parser('load',
-                                        help='Open IPython terminal with a '
-                                        'given device loaded.')
-    parser_load.add_argument('device_names', nargs='+',
-                             help='The names of one or more devices to load')
-    parser_update = subparsers.add_parser("update",
-                                          help="Update happi db "
-                                          "with JSON payload.")
-    parser_update.add_argument("json", help="JSON payload.",
-                               default="-", nargs="*")
-    parser_update = subparsers.add_parser("container-registry",
-                                          help="Print container registry.")
-    parser_transfer = subparsers.add_parser(
-                        "transfer", help="change the container of an item.")
-    parser_transfer.add_argument("name", help="Name of the item to edit")
-    parser_transfer.add_argument("target",
-                                 help="Container to transfer item into")
-    return parser
+version_msg = f'Happi: Version {happi.__version__} from {happi.__file__}'
 
 
-def happi_cli(args):
-    parser = get_parser()
-    # print happi usage if no arguments are provided
-    if not args:
-        parser.print_usage()
-        return
-    args = parser.parse_args(args)
-
+@click.group()
+@click.option('--path', type=click.Path(exists=True),
+              help='Provide the path to happi configuration file.')
+@click.option('--verbose', '-v', is_flag=True,
+              help='Show the debug logging stream.')
+@click.version_option(None, '--version', '-V', message=version_msg)
+@click.pass_context
+def happi_cli(ctx, path, verbose):
+    """Happi command line tool"""
     # Logging Level handling
-    if args.verbose:
+    if verbose:
         shown_logger = logging.getLogger()
         level = "DEBUG"
     else:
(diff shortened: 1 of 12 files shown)


happi/prompt.py
```diff
@@ -1,12 +1,80 @@
...
+    click.echo(prompt + '\nKey must be a string.  ' +
+               'Enter a blank key to complete dict entry.')
+    while True:
+        key = click.prompt('  key',
+                           default=hopefully_unique_keyword,
+                           show_default=False,
+                           value_proc=is_valid_identifier_not_keyword)
+        if key is hopefully_unique_keyword:
+            break
+        value = click.prompt('  value')
```
klauer, line 28: `ast.literal_eval` of value as well, perhaps?

tangkong:
I think I should probably stop before this PR gets any longer and open it up for reviews.

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
+    if repeated:
+        final_results = repeated
+    elif regex_list and not is_range:
+        # only matched with search_regex()
+        final_results = regex_list
+    elif range_list and is_range:
+        # only matched with search_range()
+        final_results = range_list
+    else:
+        final_results = []
```
klauer, line 137: I'd like to revisit all of this... Out of scope of this PR

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
+
+    if json:
+        json.dump([dict(res.item) for res in final_results], indent=2,
```
klauer, line 140: `json` support is busted; the import is shadowed by the variable.
Seems like it needs coverage in the test suite (even just a smoke test)

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
+            logger.info(f'Invalid device container {response}')
+            return
+
+    container = registry[response]
+    logger.debug(f'Contaner selected: {container.__name__}')
+
+    # Collect values for each field
+    kwargs = {}
+    for info in container.entry_info:
+        item_value = prompt_for_entry(info, clone_source=clone_source)
```
klauer, line 194: I entered an invalid value for a `float` and `happi add` exited on me:
```
Enter value for z, enforce=float [-1.0]: z
...
  File "/Users/klauer/Repos/happi/happi/prompt.py", line 76, in prompt_for_entry
    value = click.prompt(val_prompt, default=default,
  File "/Users/klauer/mc/envs/lucid/lib/python3.9/site-packages/click/termui.py", line 175, in prompt
    result = value_proc(value)
ValueError: could not convert string to float: 'z'
(lucid) PC98125:happi klauer$

```

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
+        clone_source = None
+        # Keep Device at registry for backwards compatibility but filter
+        # it out of new devices options
+        options = os.linesep.join(
+            [k for k, _ in registry.items() if k != "Device"]
+        )
+        ctnr_prompt = (
+            'Please select a container, or press enter for generic '
+            f'Ophyd Device container: {os.linesep}{options}'
+            '{}{}Selection'.format(os.linesep, os.linesep)
```
klauer, line 180: Mixing f-strings and `.format()`?

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
-            items_input = json.load(sys.stdin)
+    else:
+        logger.info('Aborting')
+
+
+@happi_cli.command()
+@click.argument('name')
+@click.argument('edits', nargs=-1, type=str)
+@click.pass_context
+def edit(ctx, name, edits):
```
klauer, line 216: I naively expected a question-prompting interface for `happi edit` with the refactor.

That _definitely_ does not have to be added here, but maybe when the user just does `edit itemname`:
```
$ happi edit at2l0
Please confirm the device info is correct [y/N]: y
[2022-05-25 16:07:05] - INFO -  Editing device
[2022-05-25 16:07:05] - INFO -  Adding / Modifying information for at2l0 ...
```
It should say "hey, uh, you didn't say what to edit"

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
+    logger.debug('Starting edit block')
+    device = client.find_device(name=name)
+    for edit in edits:
+        field, value = edit.split('=', 1)
+        # validate field
+        try:
+            field = is_valid_identifier_not_keyword(field)
+        except ValueError:
+            logger.error(f'field ({field}) is not a valid keyword')
+            raise click.Abort()
```
klauer, line 233: What's the difference of `click.Abort` and `click.UsageError` which I've seen used elsewhere here?

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
+
+        # validate field depending on enforce type
+        if device._info_attrs[field].enforce in (dict, list):
```
klauer, line 236: Now this is a long-standing bugfix!

I think you could also do `happi edit item 'dict={"...": "..."}'` to avoid having the shell get in the way of quote interpretation (note the single quotes surrounding)

setup.py
```diff
@@ -13,4 +13,4 @@
       include_package_data=True,
       install_requires=requirements,
       description='Happi Database Access for LCLS Beamline Devices',
-      entry_points={'console_scripts': ['happi=happi.cli:main']})
+      entry_points={'console_scripts': ['happi=happi.cli:happi_cli']})
```
ZLLentz, line 16: This rename will force me to finally update the installs in the dev environment

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
+        raise click.Abort()
+
+    for name in names:
+        devices[name] = client.load_device(name=name)
+
+    from IPython import start_ipython  # noqa
+    start_ipython(argv=['--quick'], user_ns=devices)
+
+
+# TODO: FIgure out how to deal with json and click.  list of args doesn't
```
ZLLentz, line 293: What exactly doesn't work here?

ZLLentz (commented):
I think this is looking clean- I'm liking the parser formulation as done with click. I'll take it for a spin now.

happi/cli.py
```diff
@@ -1,93 +1,43 @@
 """
```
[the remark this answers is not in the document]
tangkong: This is actually easy enough to add.  Well, at least there's a bit of documentation on how to.

happi/cli.py
```diff
@@ -1,93 +1,43 @@
...
 This module defines the ``happi`` command line interface.
 """
-import argparse
+import ast
 import fnmatch
 import json
 import logging
 import os
+# allows arrow key navigation in prompt
```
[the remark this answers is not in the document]
tangkong: The import isn't used anywhere, it fixes the issue just by being imported.  I'm ok taking out that comment, but thought it'd be confusing otherwise

happi/cli.py
```diff
@@ -1,93 +1,43 @@
...
+import readline  # noqa
```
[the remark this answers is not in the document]
tangkong: The 'check-if-import-is-unused' check.  (as above)

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
-    client = happi.client.Client.from_config(cfg=args.path)
+    # gather client
+    client = happi.client.Client.from_config(cfg=path)
     logger.debug("Happi client: %r" % client)
-    logger.debug('Happi command: %r' % args.cmd)
 
-    if args.cmd == 'search':
-        logger.debug("We're in the search block")
+    # insert items into context to be passed to subcommands
+    ctx.obj = client
```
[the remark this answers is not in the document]
tangkong: The `ctx.obj` attribute is special in `click`.  It exists on the context object as packaged, and is meant to hold whatever extra object/state you want to pass around.  I'm not sure why you can't specify your own attributes (suspect it's because context objects are created and destroyed, and obj is passed along)

setup.py
tangkong: I could change it back it's unwanted work, but I didn't have any issues invoking `happi`, so I left that way.

happi/cli.py
tangkong: It's possible this works now, I was worried about json inputs at some point.  Maybe it works now, tests seem to pass.  I'll revisit

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
+    elif len(target_match) != 1:
+        print(f'Target container name ({target}) not specific enough')
+        raise click.Abort()
+
+    target = happi.containers.registry._registry[target_match[0]]
+    # transfer item and prompt for fixes
+    transfer_container(client, item, target)
 
 
 def main():
```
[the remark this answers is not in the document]
tangkong: Done!

happi/cli.py
tangkong: This was definitely a "tests pass moving on" moment for me 😬

happi/cli.py
tangkong: Slight differences in what kinds of exceptions `click.prompt` catches lead to this.  `UsageError` is the only exception that can be raised and continue prompting.  I had our `EnforceError` subclass `UsageError`, but I didn't fully cover our enforce_value function in my previous enforce_doc work.  

This also led to fixes in the regex handling area. 👍

happi/cli.py
tangkong: oops I must have gone blind.

happi/cli.py
tangkong: For the purposes of this PR, `click.UsageError` is caught in `click.prompt`, which will re-prompt.  `click.Abort` is a RuntimeError that signals to click to terminate execution and output 1 to stderr.

happi/cli.py
tangkong: That's right, dealing with quotes is something I was mulling over for a while.  I'm not sure which way users would like more (extra quotes or escaping quotes), but I guess that'll shake out from here.

happi/cli.py
tangkong: aaaaah whoops

happi/cli.py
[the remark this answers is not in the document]
klauer: Interesting that it's on-import behavior.

I wish there were some explicit call we could make here to enable the functionality instead. Docs: https://docs.python.org/3/library/readline.html
Nothing relevant we could use/set as far as I can see.

happi/cli.py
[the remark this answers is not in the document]
klauer: Interesting - I think for our future selves it might be worthwhile making what you just said a comment about `obj`?

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
-            try:
-                getattr(device, field)
-                logger.info('Setting %s.%s = %s', args.name, field, value)
-                setattr(device, field, value)
-            except Exception as e:
-                is_invalid_field = True
-                logger.error('Could not edit %s.%s: %s', args.name, field, e)
-        if is_invalid_field:
-            sys.exit(1)
+            client_args[criteria] = fnmatch.translate(value)
```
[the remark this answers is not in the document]
tangkong: I think I'll add this to a follow up issue.  `happi search` needs to be looked at in general

happi/cli.py
tangkong: I need to add tests for this in general, maybe tests also go in a new PR.  (I have a bunch of stashed changes fixing deprecation warnings for happi.device.Device)

happi/cli.py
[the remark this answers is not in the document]
tangkong: Added to #250

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
-
-        device = client.create_device(container, **kwargs)
-        logger.info('Please confirm the following info is correct:')
-        device.show_info()
-        ok = input('y/N\n')
-        if 'y' in ok:
-            logger.info('Adding device')
-            device.save()
+            return
+        if value.replace('.', '').isnumeric():
```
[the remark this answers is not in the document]
tangkong: Noted in #250

happi/cli.py
```diff
@@ -1,93 +1,43 @@
...
 import fnmatch
 import json
 import logging
 import os
+# allows arrow key navigation in prompt
+import readline  # noqa
 import sys
 
+import click
 import coloredlogs
```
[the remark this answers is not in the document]
tangkong: Issue created #251

happi/cli.py
tangkong: Added a quick message and abort clause in 
ec48473.  The interactive prompt is a good idea though, I'll make an issue for it.

happi/cli.py
```diff
@@ -97,223 +47,314 @@ def happi_cli(args):
...
+            logger.debug('Changed %s to float', value)
+            value = str(float(value))
+
+        if is_a_range(value):
+            start, stop = value.split(',')
+            start = float(start)
+            stop = float(stop)
+            is_range = True
+            if start < stop:
+                range_list = client.search_range(criteria, start, stop)
```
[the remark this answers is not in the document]
tangkong: added to #250

tangkong:
Alrighty, I think I've made issues for all of the outstanding issues mentioned in the discussions (thanks everyone).  

There's a lot of things we can do here, but I think we agree that we can merge this and address issues individually.

setup.py
ZLLentz: Ahh I see, the old entrypoint still exists, so it still works with just a PYTHONPATH update rather than needing a full reinstall

I don't mind either way

happi/cli.py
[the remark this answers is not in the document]
tangkong: https://docs.python.org/3/library/functions.html#input

It's actually mentioned in the `input()` docs.  Doesn't make it any better, but there it is...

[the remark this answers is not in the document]
ZLLentz: I was trying to get a feel for what all of this means and found that `python -i` implicitly imports `readline` and therefore handles `input()` differently. Bizarre behavior all-around, definitely correct to import it here...

ZLLentz (approved):

happi/cli.py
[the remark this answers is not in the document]
klauer: Weird indeed... What about changing _where_ the import happens? Would it be OK if the import were in `happi.cli.main()` (or the name-changed version, I forget)?

While I can't imagine a scenario where another package would complain about readline being imported, it seems reasonable to shove the on-import happenings in the happi CLI handler instead.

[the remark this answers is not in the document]
ZLLentz: It is definitely pretty screwed up that someone importing a package somewhere in any module somewhere could change your keybind handlings subtly in specific cases. I think generally people want the readline behavior though, thankfully.

I have no strong opinions on if moving the import matters or not.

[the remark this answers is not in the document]
tangkong: I might leave it where it is, since it is technically possible to run the commands individually without invoking the parent functions (`happi.cli.main()` or `happi.cli.happi_cli()`).  It requires building the correct context object, but it is something that can happen

klauer (approved):
mamba bugs, happi CLI wonkiness and bugs galore, and readline-functionality-on-import - this PR has it all.

LGTM 👍

Issue #250, opened 2022-05-26
`happi search` logic updates and enhancements


tangkong:
## Expected Behavior

1. Better support of both glob and regex formatting.  

_Originally posted by @klauer in https://github.com/pcdshub/happi/pull/249#discussion_r882185052_

2. Handling of numeric items is... dated 
```python
if value.replace('.', '').isnumeric():
```  

_Originally posted by @klauer in https://github.com/pcdshub/happi/pull/249#discussion_r882181261_

3. Range logic doesn't quite make sense.  
  - We're limited to only one range per search (range list is replaced by whatever the final range is)
  - The final result is `regex_list + range_list`... rather than the overlap between the two?  

_Originally posted by @ZLLentz in https://github.com/pcdshub/happi/pull/249#discussion_r882183759_
  

## Possible Solution, per problem

1. With click, it may be possible to introduce syntax like `happi search --regex .*` or `happi search --glob *`.  
2. proper type casting or ast.literal_eval could be used here?
3. more thinking to come...

## Context
The `happi search` command was passed over during the click refactor, but could use a fresh look. Tests will need to be updated accordingly.  

## Your Environment
pcds-5.4.0

tangkong:
Some notes to self: 

1: python supplies [`fnmatch`](https://docs.python.org/3/library/fnmatch.html#fnmatch.translate) for translating globs to regex 

2: [Someone](https://stackoverflow.com/questions/354038/how-do-i-check-if-a-string-is-a-number-float) did a lot of work and `value.replace('.', '', 1).isdigit` seems to be fastest

tangkong:
Largely closed by #254, followup issues made where appropriate

PR #254, opened 2022-05-31
ENH: search cli enhancements / refactor

tangkong:
## Description
- add glob/regex flag, and related documentation
- fix numeric handling 
- fix range_search logic.  (Previously would only consider last range query, will now properly consider many ranges)
- adds / modifies test cases to better cover search behavior (search range intersections, exit codes, regex vs glob, etc)
- If input is invalid we exit with exit code 1 via `click.Abort()` (bad search range).  If search simply fails to return anything, the program was successful.  
- Various logic modifications to make the diff annoying

## Motivation and Context
#250 , but a more careful look is ongoing

## How Has This Been Tested?
Interactively, and tests to come

This PR.

image

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -17,7 +17,7 @@ import prettytable
 import happi
 
 from .prompt import prompt_for_entry, transfer_container
-from .utils import is_a_range, is_valid_identifier_not_keyword
+from .utils import is_a_range, is_number, is_valid_identifier_not_keyword
 
 logger = logging.getLogger(__name__)
 
@@ -96,8 +96,8 @@ def search(ctx, show_json, names, use_glob, search_criteria):
                 'Received duplicate search criteria %s=%r (was %r)',
                 criteria, value, client_args[criteria]
             )
-            return
-        if value.replace('.', '', 1).isdigit():
+            raise click.Abort()
+        if is_number(value):
             logger.debug('Changed %s to float', value)
             value = str(float(value))
 
@@ -110,6 +110,7 @@ def search(ctx, show_json, names, use_glob, search_criteria):
                 range_list += client.search_range(criteria, start, stop)
             else:
                 logger.error('Invalid range, make sure start < stop')
+                raise click.Abort()
 
             continue
 
@@ -143,6 +144,8 @@ def search(ctx, show_json, names, use_glob, search_criteria):
         final_results = range_list
     else:
         final_results = []
+        logger.error('No devices found')
+        return
 
     if show_json:
         json.dump([dict(res.item) for res in final_results], indent=2,
@@ -154,8 +157,6 @@ def search(ctx, show_json, names, use_glob, search_criteria):
         for res in final_results:
             res.item.show_info()
 
-    if not final_results:
-        logger.error('No devices found')
     return final_results
 
 
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -64,9 +64,12 @@ def happi_cli(ctx, path, verbose):
               help='Show results in JSON format.')
 @click.option('--names', is_flag=True,
               help='Return results as whitespace-separated names.')
+@click.option('--glob/--regex', 'use_glob', default=True,
+              help='use glob style (default) or regex style search terms. '
+              'Regex requires backslashes to be escaped (eg. at\\\\d.\\\\d)')
 @click.argument('search_criteria', nargs=-1)
 @click.pass_context
-def search(ctx, show_json, names, search_criteria):
+def search(ctx, show_json, names, use_glob, search_criteria):
     """
     Search the happi database.  SEARCH_CRITERIA take the form: field=value.
     If 'field=' is omitted, it will assumed to be 'name'.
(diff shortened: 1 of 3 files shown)


happi/cli.py
```diff
@@ -94,53 +96,68 @@ def search(ctx, show_json, names, search_criteria):
...
+
+            if not range_found:
+                # if first range, just replace
+                range_found = True
+                range_list = new_range_list
+            else:
+                # subsequent ranges, only take intersection
+                final_range = [item for item in new_range_list
+                               if item in range_list]
+                range_list = final_range
```
ZLLentz, line 119: This is totally good and probably doesn't need to change, but you should know that the `set` type in python makes this trivial via the `intersection` method. There are probably also other parts of this implementation that are simplified by using sets instead of lists.
https://docs.python.org/3.10/library/stdtypes.html#frozenset.intersection

ZLLentz (approved):
I haven't taken this for a spin yet, might come back to it after clearing the rest of the queue, but this looks good to me. I added one small suggestion about leveraging the `set` methods that doesn't need to be followed if it ends up not making sense.

happi/cli.py
tangkong: I thought about using sets, but the SearchResult items aren't hashable so we'd have to grab the id's, do the set operations, and go back to the search results.  

I figured that was more effort than the list comprehension, and from what I can tell it's not too much slower

klauer: Hmm, we could make `SearchResult` hashable - I think that'd be a good improvement regardless of the implementation here

happi/cli.py
```diff
@@ -64,9 +64,12 @@ def happi_cli(ctx, path, verbose):
               help='Show results in JSON format.')
 @click.option('--names', is_flag=True,
               help='Return results as whitespace-separated names.')
+@click.option('--glob/--regex', 'use_glob', default=True,
+              help='use glob style (default) or regex style search terms. '
+              'Regex requires backslashes to be escaped (eg. at\\\\d.\\\\d)')
```
klauer, line 69: Raw string to lessen the escape chars, perhaps?
```suggestion
              r"Regex requires backslashes to be escaped (eg. at\d.\d)"
```

klauer: Also I like that `click.option` can easily specify the mutually-exclusive flags here. Neat!

happi/cli.py
```diff
@@ -64,9 +64,12 @@ def happi_cli(ctx, path, verbose):
...
 @click.argument('search_criteria', nargs=-1)
 @click.pass_context
-def search(ctx, show_json, names, search_criteria):
+def search(ctx, show_json, names, use_glob, search_criteria):
```
klauer, line 72: Time to toss in some annotations, perhaps?

happi/cli.py
```diff
@@ -94,53 +96,68 @@ def search(ctx, show_json, names, search_criteria):
...
+
+            continue
 
-        # skip the criteria for range values
-        # it won't be a valid criteria for search_regex()
-        if is_range:
-            pass
+        elif is_number(value):
+            logger.debug('Changed %s to float', value)
+            value = str(float(value))
```
klauer, line 131: Time to bring this up again - in your digging, do you understand why we go from `str` -> `float` -> `str`?

happi/cli.py
```diff
@@ -94,53 +96,68 @@ def search(ctx, show_json, names, search_criteria):
                 'Received duplicate search criteria %s=%r (was %r)',
                 criteria, value, client_args[criteria]
             )
-            return
-        if value.replace('.', '').isnumeric():
-            logger.debug('Changed %s to float', value)
-            value = str(float(value))
+            raise click.Abort()
 
         if is_a_range(value):
```
klauer, line 101: Not a required change but a refactor thought: it would be nice for reusability's sake to say "give me a CLI argument (as the user would format a key/value pair or a range) and turn it into a dictionary of query parameters"

klauer (approved):
I think this looks like a big improvement!

See below for my usual grain-of-salt suggestions and thoughts.

happi/cli.py
tangkong: It's because these arguments get passed to `client.search_regex`, which requires strings.  I assume we went `str` -> `float` to make sure the number was formatted correctly, but as I think about it more deeply, I wonder if this works for EntryItems that require `int`s.  I'll look into this some more.

happi/cli.py
tangkong: I thought about this briefly, and the other place we use this kind of parsing is in `happi edit`.  The similarities end around the `arg.split('=', 1)` point, so I didn't think it was worth splitting it out at that point.  But it would certainly make the search block more readable.

happi/cli.py
tangkong: Ah yes if we have some value that's stored as an int rather than a float, the search_regex method won't work at all.  I'm willing to bet we've never run into this because our only numeric field is the float field `z`

tangkong: For now I'll leave this in.  For our current use case, it allows users to search `z=735` and still match `z=735.0`.  It's something that will need to be looked at in the future, however.

klauer: Sounds good. I'm glad you dug into those details - and it's worth an issue/follow-up at some point for sure.

PR #257, opened 2022-06-03
DOC: write release notes for v1.13.0

ZLLentz:
big tag


tangkong (approved):
LGTM.  I think adding pre-release notes might be in order if I keep working on `happi`

Issue #255, opened 2022-06-02
Make `happi.SearchResult` hashable for usage in sets and such


klauer:
_Originally posted by @klauer in https://github.com/pcdshub/happi/pull/254#discussion_r888369082_

tangkong:
A thought here: Typically you only want to hash immutable objects, lest their hashes change and they get lost in the hash table somewhere.  The use case we are considering right now (finding intersections of sets of `SearchResult`s) is innocuous enough, but a `SearchResult` is very much a mutable object.  

Perhaps it shouldn't be though.  If we're modifying a search result before instantiating an object or something similar, there's a problem

klauer:
I think `SearchResult` isn't really intentionally mutable - just it has a dictionary of metadata that you could come along and modify if you so pleased.

Since each result pairs with a given identifier from the database, the hash key of a `SearchResult` could be something like `(client, metadata["_id"])`

tangkong:
I have a working implementation that bases the hash on the metadata dictionary, which I was considering making a `frozendict` to alleviate my mutability concerns.  I figure if we don't intend the object to be mutable we may as well force it to be as such

PR #260, opened 2022-06-07
ENH: Hashable SearchResult

tangkong:
## Description
- Implements `__eq__` and `__hash__` to make `SearchResult` hashable
- uses above hashability in `happi search` command

## Motivation and Context
We used to manual set intersections, but now we can do them more concselt
#255 

## How Has This Been Tested?
Test Suite additions, this PR

This PR

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -121,8 +121,7 @@ def search(
                 range_list = new_range_list
             else:
                 # subsequent ranges, only take intersection
-                final_range = [item for item in new_range_list
-                               if item in range_list]
+                final_range = set(new_range_list) & set(range_list)
                 range_list = final_range
 
             if not range_list:
@@ -156,7 +155,7 @@ def search(
         final_results = range_list
     elif range_list and regex_list:
         # find the intersection between regex_list and range_list
-        final_results = [item for item in range_list if item in regex_list]
+        final_results = set(range_list) & set(regex_list)
     else:
         logger.debug('No regex or range items found')
 
--- a/happi/tests/test_client.py
+++ b/happi/tests/test_client.py
@@ -316,8 +316,24 @@ def test_choices_for_field(happi_client: Client):
 
 def test_searchresults(client_with_three_valves: Client):
     valve1 = client_with_three_valves['VALVE1']
-    assert valve1.metadata == valve1
     assert isinstance(valve1.get(), types.SimpleNamespace)
+
+
+def test_hashable_searchresults(client_with_three_valves: Client):
+    client = client_with_three_valves
+    valve1 = client['VALVE1']
+    valve2 = client['VALVE2']
+    valve3 = client['VALVE3']
+    results1 = set(client.search_regex(name='valve1|valve2'))
+    results2 = set(client.search_regex(name='valve.*'))
+
+    assert len(results1 & results2) == 2
+    assert valve1 in (results1 & results2)
+    assert valve2 in (results1 & results2)
+    assert len(results1 ^ results2) == 1
+    assert valve3 in (results1 ^ results2)
+    assert valve2 not in (results1 ^ results2)
+    assert results2 == (results1 | results2)
 
 
 def test_client_mapping(
--- a/happi/client.py
+++ b/happi/client.py
@@ -2,6 +2,7 @@ import collections
 import configparser
 import inspect
 import itertools
+import json
 import logging
 import os
 import re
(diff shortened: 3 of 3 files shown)


klauer:
I'm a big 👍 on the overall hashability/equality checking functionality, but I think we should take a moment to discuss and come to an agreement on what the hash should actually do for us - and then agree on the hash key itself.

`__hash__` in this PR says: "use every piece of the metadata to make a hash key!". This is as strict as a hash key that could possibly make sense. Anything additional would overconstrain it in undesirable ways.

Is using the entirety of the metadata really something we want? Let's consider:
* If I perform two searches that yield the same `HappiItem` with the same metadata, their hashes will match as expected
* If I perform two searches that yield the same `HappiItem` with the differing metadata (e.g., the `z` location changed externally), their hashes will differ

In the latter scenario, this would mean a `set()` could contain two of the same `HappiItem`, one with different metadata than the other. `SearchResult` could not be used as a dictionary/mapping key effectively here either. I believe this is not what we want.

My take on it is that what we care about when hashing a `SearchResult` (and therefore `HappiItem`) is as follows:
1. The database from which the item comes from. That is, using a client that talks to one database and a client that talks to another database, `client1.search()` and `client2.search()` should never yield the same hash
2. The unique key for the item (i.e., `_id_key` or pretty much the `name` kwarg)

With the above, no matter if the metadata changes between searches, the hashes would remain the same.

tangkong:
Good points, I suppose I approached the problem from a different direction.  

I presumed that what we cared about in the end was the device itself.  That is, even if we had two results from different databases, if their metadata was the same they would effectively be the same device and produce the same ophyd object. 
  I suppose my main thought is that at least the metadata used to instantiate the item should be used to determine if it is unique.

In the case where we perform multiple searches and get the same item with differing metadata, isn't that a case where we should be questioning what's going on with the client/database in the first place?

As for hashing based on the `client` and `_id`:
- right now clients aren't hashable either, though we could fix that I'm sure
- I know `_id` is supposed to be a unique identifier in a `happi` database, but thus far haven't seen any guarantees of this . Part of this is due to us not doing any validation on the database, except on insertion.

klauer:
> I presumed that what we cared about in the end was the device itself. That is, even if we had two results from different databases, if their metadata was the same they would effectively be the same device and produce the same ophyd object.
I suppose my main thought is that at least the metadata used to instantiate the item should be used to determine if it is unique.

That's a fair point. 

I'm not sure that happi can make guarantees about the object that comes out of `.get()`. I think its concern stops at the database level and not what you get back when you try to instantiate an object.

> In the case where we perform multiple searches and get the same item with differing metadata, isn't that a case where we should be questioning what's going on with the client/database in the first place?

Not necessarily - there's nothing stopping people from editing the database as you search through it or read it. I think the `z` update above is a pretty simple - and likely - example of that scenario.

Searches could happen at different times during the life of a GUI program, and I'd still want "one unique identifier in the database" to always hash to the same thing.

>  I know _id is supposed to be a unique identifier in a happi database, but thus far haven't seen any guarantees of this . Part of this is due to us not doing any validation on the database, except on insertion.

It's done somewhat implicitly. The JSON backend wouldn't allow two keys with the same name, so the uniqueness is built-in to the file format itself. 

I don't recall the specifics for the mongo backend, to be honest. I believe the intent has always been that the id should be unique to the database.

ZLLentz:
This is pretty true to the existing behavior of comparing search results. Ken makes some pretty good points about the practicality of hashing searches that were made far apart in time if the metadata has been edited- revealing a small race condition that already exists in the code prior to this full md hashing approach.

I'm surprised that the search result contains the full md raw rather than a database reference of some kind that can be used to load the md- that is, I'm surprised that the metadata at this point in time is so strongly linked to the result of searching. It's probably not worth thinking too hard about alternate ways to frame this.

I think there's some innate conflict here between what a `SearchResult` sounds like it should be (here's some ids that match your criteria) vs what the implementation says that it is (here's all the md dicts that have keys that match your search criteria).

Robert's implementation here matches what the class actually is, Ken's suggestion matches the use case outlined in the CLI and probably the most common use case in general. I'm not sure what is correct even after thinking for a bit.

tangkong:
I think I'm ok with hashing on the client and unique `_id`, with the knowledge that we'll have to deal with metadata conflicts / edited data in the future.

happi/tests/test_client.py
```diff
@@ -316,10 +316,29 @@ def test_choices_for_field(happi_client: Client):
 
 def test_searchresults(client_with_three_valves: Client):
     valve1 = client_with_three_valves['VALVE1']
-    assert valve1.metadata == valve1
```
klauer, line 319: Could this be brought back?

klauer (approved):
Not much else to say - because I already said a lot - LGTM 👍

happi/tests/test_client.py
tangkong: This assert clause should fail, since the `valve1` and `valve1.metadata` aren't the same class.  Is this behavior something we want?

klauer: Seems the original intent was: "did I get back what I requested?" Even a minimal check would suffice, even if not all the metadata

happi/cli.py
```diff
@@ -121,8 +121,7 @@ def search(
                 range_list = new_range_list
             else:
                 # subsequent ranges, only take intersection
-                final_range = [item for item in new_range_list
-                               if item in range_list]
+                final_range = set(new_range_list) & set(range_list)
                 range_list = final_range
```
ZLLentz, line 125: Probably should have caught this last PR: this bit can be dropped down to one line, right? I don't think `final_range` is ever used, except to be set here.

ZLLentz (approved):
Looking good

PR #261, opened 2022-06-08
BUG: allow `happi search` to match int numerics

### LINKED ISSUE #256: Handle searching numeric values
If we have some value that's stored as an int rather than a float, the search_regex method won't match the item.  In all likelihood we've never run into this because our only numeric field is the float field `z`

_Originally posted by @tangkong in https://github.com/pcdshub/happi/pull/254#discussion_r888431458_

This is in some sense a limitation of `client.search_regex`, perhaps there's a better way to search for numeric values that will require them being separated from the rest of the search terms

tangkong:
## Description
- If search term is an int, generates regex that also matches its float counterpart

## Motivation and Context
closes #256 

In building tests I realized `happi edit` does not allow editing of `extraneous` information.  A portion of the new test has been commented out for when that is fixed.  (see #262)

## How Has This Been Tested?
Test suite, interactively

This PR

image

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -132,8 +132,16 @@ def search(
             continue
 
         elif is_number(value):
-            logger.debug('Changed %s to float', value)
-            # value = str(float(value))
+            if float(value) == int(float(value)):
+                # value is an int, allow the float version (optional .0)
+                logger.debug(f'looking for int value: {value}')
+                value = f'^{int(float(value))}(\\.0+$)?$'
+
+                # don't translate from glob
+                client_args[criteria] = value
+                continue
+            else:
+                value = str(float(value))
         else:
             logger.debug('Value %s interpreted as string', value)
 
--- a/happi/tests/test_cli.py
+++ b/happi/tests/test_cli.py
@@ -219,6 +219,29 @@ def test_search_z_range(
     assert 'No devices found' in conflict_result.output
 
 
+def test_search_int_float(runner: CliRunner, happi_cfg: str):
+    int_result = runner.invoke(happi_cli, ['--path', happi_cfg,
+                               'search', '--names', 'z=3'])
+    float_result = runner.invoke(happi_cli, ['--path', happi_cfg,
+                                 'search', '--names', 'z=3.0'])
+    assert int_result.output == float_result.output
+
+    # # TODO: add this test case once edit works on extraneous info
+    # edit_result = runner.invoke(
+    #     happi_cli,
+    #     ['-v', '--path', happi_cfg, 'edit', 'tst_base_pim', 'z=3.001'],
+    #     input='y'
+    # )
+    # assert edit_result.exit_code == 0
+
+    # int_result = runner.invoke(happi_cli, ['--path', happi_cfg,
+    #                            'search', '--names', 'z=3'])
+    # float_result = runner.invoke(happi_cli, ['--path', happi_cfg,
+    #                              'search', '--names', 'z=3.0'])
+    # assert int_result.output == ''
+    # assert float_result.output == ''
+
+
 def test_both_range_and_regex_search(client: happi.client.Client):
     # we're only interested in getting this entry (TST_BASE_PIM2)
     res = client.search_regex(z='6.0')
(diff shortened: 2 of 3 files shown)


happi/cli.py
```diff
@@ -132,8 +132,16 @@ def search(
...
+            if float(value) == int(float(value)):
+                # value is an int, allow the float version (optional .0)
+                logger.debug(f'looking for int value: {value}')
+                value = f'^{int(float(value))}(\\.0+$)?$'
+
+                # don't translate from glob
+                client_args[criteria] = value
+                continue
+            else:
+                value = str(float(value))
```
klauer, line 144: Couldn't a floating point value be somewhat problematic for regex mode? The above comment and continue gave me pause to think about it a bit:

* Say we passed in "3.14" we might match "3514" or "3x14" just the same. `.` would be interpreted as a regex anychar, wouldn't it?
* Does float-based value searching also work as a full match or a partial match? That is, if `12345.678` is in the database and I search for `5.678` what happens? (lazy question...)

happi/cli.py
```diff
@@ -132,8 +132,16 @@ def search(
             continue
 
         elif is_number(value):
-            logger.debug('Changed %s to float', value)
-            # value = str(float(value))
+            if float(value) == int(float(value)):
+                # value is an int, allow the float version (optional .0)
+                logger.debug(f'looking for int value: {value}')
+                value = f'^{int(float(value))}(\\.0+$)?$'
```
klauer, line 138: I think this is good as you have it: 
* Requiring it to match the entire integer value - as marked by `^...$`
* With or without the trailing `.0`.

happi/cli.py
tangkong: These are good points.  
- I think you're right, ints will always get an escaped `.`, but floats will get passed through as is.  Is this a problem though?  We could argue if someone is searching in regex mode they should be escaping their `.` if they need to?
- The float-based value searching seems to not work with partial matches, though I didn't do that intentionally.

happi/tests/test_cli.py
```diff
@@ -219,6 +219,29 @@ def test_search_z_range(
...
+    #     input='y'
+    # )
+    # assert edit_result.exit_code == 0
+
+    # int_result = runner.invoke(happi_cli, ['--path', happi_cfg,
+    #                            'search', '--names', 'z=3'])
+    # float_result = runner.invoke(happi_cli, ['--path', happi_cfg,
+    #                              'search', '--names', 'z=3.0'])
+    # assert int_result.output == ''
+    # assert float_result.output == ''
```
ZLLentz, line 242: :+1: for leaving a test for future use
I think another way to do this is to add the test and mark it as xfail

ZLLentz:
The screenshot in the description above confused the heck out of me because its "after -> before" instead of "before -> after" but I caught myself before making an equally confused comment.

ZLLentz (approved):
I think this is good

happi/cli.py
klauer: > We could argue if someone is searching in regex mode they should be escaping their . if they need to?

If someone is doing a regex search with a thing that looks like a floating point value... is it a floating point value or a regex?

> Is this a problem though? 

Likely not - just trying to dig into the implications of the changes in my usual excessive way 😁

ZLLentz: I think this is an annoyance with using regex in general to search for floating point numbers rather than a unique problem to happi

Float-based values not matching almost seems like a feature rather than a bug...

klauer (approved):
My annoying questions shouldn't hold this up. It's a net positive improvement 👍

happi/cli.py
ZLLentz: That's a good point about the processing of regex, Ken

tangkong: > Likely not - just trying to dig into the implications of the changes in my usual excessive way 😁

I don't think it's excessive, it's just thorough 😄 

>Float-based values not partial matching almost seems like a feature rather than a bug...

I wish I could claim credit for that haha

Alpha-hub1 (commented):
Hey

Issue #235, opened 2022-03-04
JSON backend can be egregiously slow on network filesystems


klauer:
This pattern was carried forward from the original implementation, just generalized: (I, uh, `git blame` shows me, but don't blame me 😬 )

Load/yield/save:
https://github.com/pcdshub/happi/blob/f8c55b2928d9b60f4a882b1ec3617051aa5a1a94/happi/backends/json_db.py#L26-L31

Load on every method:
https://github.com/pcdshub/happi/blob/f8c55b2928d9b60f4a882b1ec3617051aa5a1a94/happi/backends/json_db.py#L68

This originally was likely an attempt to get around the inability to lock files on the filesystem / concurrent access / etc - all of those things a real database is meant to help with.

What would be a reasonable alternative here - at least keeping the spirit of the original, not just saving on exit?

PR #265, opened 2022-06-13
REF/PERF: caching mechanism for happi Client and backend databases

klauer:
## Description
* Adds cache clearing mechanism to backends
* Each call to `client.search()` (and such) will hit the database once and cache internally
* User of client can further indicate that caching is to be done by way of the `retain_cache_context`
* Noted an additional performance issue: 
    * Search methods retrieve entire documents
    * When the client tries to instantiate a `HappiItem`, it hits the database again with just the identifier, retrieving the same document prior to instantiation
    * This is where the `_get_item_from_document` refactor comes in

## Motivation and Context
Addresses #235

## How Has This Been Tested?
Solely the test suite

Issue and PR

### DIFF
--- a/happi/client.py
+++ b/happi/client.py
@@ -1,5 +1,6 @@
 import collections
 import configparser
+import contextlib
 import inspect
 import itertools
 import logging
@@ -128,6 +129,8 @@ class Client(collections.abc.Mapping):
     _results_wrap_class = SearchResult
 
     def __init__(self, database=None, **kwargs):
+        self._retain_cache = False
+
         # Use supplied backend
         if database:
             self.backend = database
@@ -143,6 +146,25 @@ class Client(collections.abc.Mapping):
                 raise DatabaseError(
                     f"Failed to instantiate a {DEFAULT_BACKEND} backend"
                 ) from exc
+
+    @contextlib.contextmanager
+    def retain_cache_context(self, clear_first: bool = True):
+        """
+        Context manager which can be used to retain the happi backend cache.
+
+        Parameters
+        ----------
+        clear_first : bool, optional
+            Clear the cache before entering the block.  Defaults to True.
+        """
+        if clear_first:
+            self.backend.clear_cache()
+
+        try:
+            self._retain_cache = True
+            yield
+        finally:
+            self._retain_cache = False
 
     def find_document(self, **kwargs):
         """
@@ -175,7 +197,7 @@ class Client(collections.abc.Mapping):
         if len(kwargs) == 0:
             raise SearchError('No information pertinent to device given')
 
-        self.backend.clear_cache()
+        self._maybe_clear_cache()
         matches = list(itertools.islice(self.backend.find(kwargs), 1))
         if not matches:
             raise SearchError(
@@ -492,6 +514,11 @@ class Client(collections.abc.Mapping):
     def __len__(self):
         return len(self.all_items)
 
+    def _maybe_clear_cache(self):
+        """Clear the backend cache if not in a retain-cache block."""
+        if not self._retain_cache:
+            self.backend.clear_cache()
+
     def _get_search_results(self, items, *, wrap_cls=None):
         """
         Return search results to the user, optionally wrapping with a class.
@@ -536,7 +563,7 @@ class Client(collections.abc.Mapping):
                                               beamline='HXR')
         """
 
-        self.backend.clear_cache()
+        self._maybe_clear_cache()
         items = self.backend.find_range(key, start=start, stop=end,
                                         to_match=kwargs)
         return self._get_search_results(items)
@@ -564,7 +591,7 @@ class Client(collections.abc.Mapping):
             hxr_valves  = client.search(type='Valve', beamline='HXR')
         """
 
-        self.backend.clear_cache()
+        self._maybe_clear_cache()
         items = self.backend.find(kwargs)
         return self._get_search_results(items)
 
@@ -594,7 +621,7 @@ class Client(collections.abc.Mapping):
             three_valves = client.search_regex(_id='VALVE[123]')
         """
 
-        self.backend.clear_cache()
+        self._maybe_clear_cache()
         items = self.backend.find_regex(kwargs, flags=flags)
         return self._get_search_results(items)
 
--- a/happi/client.py
+++ b/happi/client.py
@@ -9,7 +9,7 @@ import re
 import sys
 import time as ttime
 import warnings
-from typing import Any, Dict, Optional, Type
+from typing import Any, Dict, Optional, Sequence, Type
 
 from . import containers
 from .backends import BACKENDS, DEFAULT_BACKEND
@@ -340,8 +340,6 @@ class Client(collections.abc.Mapping):
         device : :class:`.HappiItem`
             A device that matches the characteristics given.
         """
-
-        logger.debug("Gathering information about the device ...")
         return self._get_item_from_document(self.find_document(**post))
 
     def load_device(self, use_cache=True, **post):
@@ -519,19 +517,23 @@ class Client(collections.abc.Mapping):
         if not self._retain_cache:
             self.backend.clear_cache()
 
-    def _get_search_results(self, items, *, wrap_cls=None):
+    def _get_search_results(
+        self, docs: Sequence[Dict[str, Any]], *, wrap_cls: Optional[type] = None
+    ):
         """
         Return search results to the user, optionally wrapping with a class.
         """
         wrap_cls = wrap_cls or self._results_wrap_class
         results = []
-        for info in items:
+        for doc in docs:
             try:
-                result = wrap_cls(client=self, device=self.find_device(**info))
+                result = wrap_cls(client=self, device=self._get_item_from_document(doc))
                 results.append(result)
             except Exception as exc:
-                logger.warning('Entry for %s is malformed (%s). Skipping.',
-                               info['name'], exc)
+                logger.warning(
+                    "Entry for %s is malformed (%s). Skipping.",
+                    doc["name"], exc
+                )
         return results
 
     def search_range(self, key, start, end=None, **kwargs):
--- a/happi/client.py
+++ b/happi/client.py
@@ -175,6 +175,7 @@ class Client(collections.abc.Mapping):
         if len(kwargs) == 0:
             raise SearchError('No information pertinent to device given')
 
+        self.backend.clear_cache()
         matches = list(itertools.islice(self.backend.find(kwargs), 1))
         if not matches:
             raise SearchError(
@@ -268,6 +269,34 @@ class Client(collections.abc.Mapping):
         device.save = save_device
         return _id
 
+    def _get_item_from_document(self, doc: Dict[str, Any]) -> HappiItem:
+        """
+        Get a HappiItem given the database document.
+
+        Parameters
+        ----------
+        post
+            Key-value pairs of search criteria used to find the device.
+
+        Returns
+        -------
+        item : :class:`.HappiItem`
+            A HappiItem instance for the document.
+        """
+
+        logger.debug("Instantiating device based on found information ...")
+        try:
+            device = self.create_device(doc['type'], **doc)
+        except (KeyError, TypeError) as exc:
+            raise EntryError('The information relating to the device class '
+                             'has been modified to the point where the object '
+                             'can not be initialized, please load the '
+                             'corresponding document') from exc
+
+        # Add the save method to the device
+        device.save = lambda: self._store(device, insert=False)
+        return device
+
     def find_device(self, **post):
         """
         Query the database for an individual HappiItem.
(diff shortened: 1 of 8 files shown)


tangkong:
I think this is a good approach, though I might put this logic in the `_load_or_initialize()` method and leave `load()` such that it always loads the database (for the sake of that method being intuitive).  

We'd probably want something similar in the mongo backend, though we don't really use that currently right?

ZLLentz:
> We'd probably want something similar in the mongo backend, though we don't really use that currently right?

In the mongo backend, the internal API won't be of the form "read the entire database file" as in the json, it will be more directly "hey server, please give me this specific document", so I suspect that it won't have the exact same problem as here for bulk reads- but maybe there will be other issues to consider separately.

The speed problem we have here comes from the backend reading the entire file to give us one key from it.

That is to say: I think it's definitely correct to do these optimizations per-backend rather than on the client.

RE: where to put the logic, I like Robert's suggestion to keep `load` as-is for direct use (give me the whole raw database right now) and wrap/cache it in the functions that the searching logic uses to facilitate these "do lots of searches at once" cases, but I want to clarify that `load` isn't even a required function for a backend so it's OK to change the behavior. No code outside the backend will be calling `load`.

5.0s as the threshold will basically never get you in trouble (you'll "always" catch updates) while being very sufficient to cut down the db reads in a bulk chain search from thousands to exactly 1.

I'd have absolutely agreed to a merge as-is.

klauer:
OK, based on those performance numbers this will go back to draft then!

ZLLentz:
Yeah I think the deepcopy breaks the performance gains:

before:
```
Total time: 3.83411 s
File: C:\Users\zlentz\Documents\VSCode\happi\happi\backends\json_db.py
Function: _load_or_initialize at line 59

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    59                                               def _load_or_initialize(self):
    60                                                   """Load an existing database or initialize a new one."""
    61       608          0.2      0.0      0.0          try:
    62       608       3833.9      6.3    100.0              return self.load()
    63                                                   except FileNotFoundError:
    64                                                       logger.debug('Initializing new database')
    65
    66                                                   self.initialize()
    67                                                   return self.load()
```
after:
```
Total time: 70.6864 s
File: C:\Users\zlentz\Documents\VSCode\happi\happi\backends\json_db.py
Function: _load_or_initialize at line 78

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    78                                               def _load_or_initialize(self):
    79                                                   """Load an existing database or initialize a new one."""
    80       608          0.3      0.0      0.0          cache_invalid = (
    81       608          0.5      0.0      0.0              self._last_loaded is None
    82       607          1.7      0.0      0.0              or time.monotonic() - self._last_loaded >= self.cache_seconds
    83                                                   )
    84
    85       608          0.3      0.0      0.0          if cache_invalid:
    86        15          0.0      0.0      0.0              try:
    87        15        146.0      9.7      0.2                  self._load_cache = self.load()
    88                                                       except FileNotFoundError:
    89                                                           logger.debug("Initializing new database")
    90                                                           self.initialize()
    91                                                           self._load_cache = self.load()
    92        15          0.1      0.0      0.0              self._last_loaded = time.monotonic()
    93
    94       608      70537.5    116.0     99.8          return copy.deepcopy(self._load_cache)
```

klauer:
I think instead we could cache the raw JSON and reload that. Let me give that a try.

klauer:
Locally:
```
In [9]: %timeit json.loads(client.backend._raw_json_cache)
3.01 ms ± 251 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
So we at least cut out the I/O operation and just require json to reprocess things - which is generally pretty quick.

When you get a chance, this look any better @ZLLentz ?

ZLLentz:
25% improvement over the original
```
Total time: 3.01955 s
File: C:\Users\zlentz\Documents\VSCode\happi\happi\backends\json_db.py
Function: _load_or_initialize at line 86

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    86                                               def _load_or_initialize(self):
    87                                                   """Load an existing database or initialize a new one."""
    88       608          2.4      0.0      0.1          if self.cache_is_valid and self._raw_json_cache is not None:
    89       607       3009.5      5.0     99.7              return self._load_from_json(self._raw_json_cache)
    90
    91         1          0.0      0.0      0.0          try:
    92         1          7.7      7.7      0.3              loaded = self.load()
    93                                                   except FileNotFoundError:
    94                                                       logger.debug("Initializing new database")
    95                                                       self.initialize()
    96                                                       loaded = self.load()
    97         1          0.0      0.0      0.0          return loaded
```

klauer:
That's a shame - is it even worth the additional complexity for 25% speed? I'm leaning toward "not really"

ZLLentz:
> For each load, a deep copy is required to keep the functionality the same as before without further modifying the client and/or tests.

What internals are getting in the way here? I think if there was a way to only make the dict once we'd save even more time, making the complexity worthwhile.

ZLLentz:
I see, totally bizarre that a test is mutating the dictionary- we can work around that probably. The further implications of caching a mutable dict are more interesting and need some thought:
1. Is there any case where a consumer is using a backend directly? Note that `load` isn't part of the backend API that `Client` uses, so the dictionary should really only be passed around within the json backend.
2. Is making the dictionary immutable a sensible thing to do? Well, maybe. More thought...

klauer:
(I reverted the last 2 commits and fixed the test suite keeping the cached result as mutable/shared. I think the main target of this refactor should be a large speedup, and it won't come without addressing the things we're discussing now.)

> * Is there any case where a consumer is using a backend directly? Note that `load` isn't part of the backend API that `Client` uses, so the dictionary should really only be passed around within the json backend.

They _shouldn't_ be, that's for sure. But we don't go to great lengths to hide the backend either (`client.backend` takes you right to it).

If we are to keep the cache and want to keep queries/invocations isolated from one another, I think the following (non-exhaustive list) would be good:
1. a `copy.deepcopy` for the `client.find_document` return value
2. `copy.deepcopy` for any `backend.find()` calls that filter back to the user (there's a few spots)

> * Is making the dictionary immutable a sensible thing to do? Well, maybe. More thought...

I think that could be a good thing - and faster than the `deepcopy` stuff above. 
It'd also be a potentially breaking API change...

ZLLentz:
- I agree that user-facing client methods should never return the cached dict and that a deepcopy is appropriate (this would happen once per call at most at the very end of a find, not repeatedly as we go, right?)
- Perhaps the easiest thing to do is to completely leave `load` as-is and add a `_cached_load` method that is used internally in the json backend

tangkong:
A really naive thought: did we try `functools.lru_cache` ?   A really basic test:

```python
In [7]: @lru_cache
   ...: def get_json(path):
   ...:     with open(path, 'r') as f:
   ...:         return json.load(f)
   ...:

In [8]: %%timeit -n 1
   ...: db = get_json(Path('/cds/home/r/roberttk/test/db.json'))
   ...:
The slowest run took 643.76 times longer than the fastest. This could mean that an intermediate result is being cached.
1.13 ms ± 2.71 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

running `happi profile -d`
```
Total time: 0.00982 s
File: /cds/home/r/roberttk/devrepos/happi/happi/backends/json_db.py
Function: _load_or_initialize at line 60

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    60                                               def _load_or_initialize(self):
    61                                                   """Load an existing database or initialize a new one."""
    62       598          0.3      0.0      3.2          try:
    63       598          9.5      0.0     96.8              return self.load()
    64                                                   except FileNotFoundError:
    65                                                       logger.debug('Initializing new database')
    66                                           
    67                                                   self.initialize()
    68                                                   return self.load()
```

I'm not sure if this is what zach was doing to benchmark things.

klauer:
> but maybe `cachetools.func.ttl_cache`?

I think the caching mechanism (whether hand-written as I had it a few commits back or imported from the standard library/some 3rd party library) isn't that important - or at the core of what we need to target here.

Recapping what we know:
1. We can cache the raw JSON contents and call `json.loads()` on them for each `backend.load()` call for a 25% speed increase
2. We can cache the deserialized JSON dictionary of the database and do a `deepcopy()` for a performance _penalty_ (in the case of our large database)
3. We can cache the deserialized JSON dictionary of the database for a massive performance boost
    a. Loads in effectively constant time regardless of the number of calls made within a 5 second period
    b. Caveat: passing back to the client/user the same mutable dictionary

I think we've largely come to the conclusion that we want the backend to be able to do (3), but we need to resolve how to deal with 3b in a way that is agreeable to existing happi consumers and the test suite.

ZLLentz:
We may also consider some combined search function on `Client` where we can chain `search`, `search_regex`, and `search_range` on the same cache rather than implementing this in downstream applications or in the cli layer as is done now

klauer:
Updated the PR text based on the discussion + proposed changes. Marking as ready for review, though there may be more to touch on here.

happi/client.py
```diff
@@ -482,6 +514,11 @@ def __iter__(self):
     def __len__(self):
         return len(self.all_items)
 
+    def _maybe_clear_cache(self):
+        """Clear the backend cache if not in a retain-cache block."""
+        if not self._retain_cache:
+            self.backend.clear_cache()
```
tangkong, line 520: Did we still want a timeout condition here?  As it is now I think the cache will never clear if we set the client to retain the cache?

happi/tests/test_client.py
```diff
@@ -351,3 +351,36 @@ def test_client_mapping(
...
+        results1 = set(client.search_regex(name='valve1|valve2'))
+        results2 = set(client.search_regex(name='valve.*'))
+
+    assert cache_clear_count == 1
+
+    assert valve1 != 'VALVE1'
+    assert valve1 != valve1.metadata
+
+    assert len(results1 & results2) == 2
+    assert valve1 in (results1 & results2)
```
tangkong, line 381: These test conditions are from the `hashable_search_results` case right?  Should they go in this test?  If I saw `test_cache_context` fail I wouldn't expect it to be due to set operations.

tangkong (approved):
This looks good to me.  Just a couple thoughts and I think this is good to merge.

happi/tests/test_client.py
klauer: Yes, it's a copy/paste of that test. My thought was "let's test that the cache loads only once and performs identically to the original test".

happi/client.py
klauer: The timeout is no longer necessary as the basis for clearing the cache is "did the user ask to clear the cache?"

The control that the user can leverage is only by way of:
1. Calling `search()` et al on their own (cache clearing on every call)
2. Using the `retain_cache_context` (cache clearing at most once per context)

klauer:
Anecdotally, with the last commit ([b822321](https://github.com/pcdshub/happi/pull/265/commits/b8223214c0e8fa0717dcc30ab26f80bc4ca51b45)) happi nearly loads instantaneously on atef. I'm officially 100% more excited for this.

ZLLentz:
Here's the same profile as above:
```
Total time: 0.0112806 s
File: C:\Users\zlentz\Documents\VSCode\happi\happi\backends\json_db.py
Function: _load_or_initialize at line 65

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    65                                               def _load_or_initialize(self):
    66                                                   """Load an existing database or initialize a new one."""
    67         1          0.0      0.0      0.0          if self._load_cache is None:
    68         1          0.0      0.0      0.0              try:
    69         1         11.3     11.3    100.0                  self._load_cache = self.load()
    70                                                       except FileNotFoundError:
    71                                                           logger.debug("Initializing new database")
    72                                                           self.initialize()
    73                                                           self._load_cache = self.load()
    74
    75         1          0.0      0.0      0.0          return self._load_cache
```
A staggering speedup that isn't helpfully expressed by a percentage!

ZLLentz:
It looks to me like the remaining db load time for the json db is largely spent in assembling the various `HappiItem` and the associated `EntryInfo` objects- not much low hanging fruit left for a json db with <1000 entries!

happi/client.py
```diff
@@ -144,6 +147,25 @@ def __init__(self, database=None, **kwargs):
                     f"Failed to instantiate a {DEFAULT_BACKEND} backend"
                 ) from exc
 
+    @contextlib.contextmanager
+    def retain_cache_context(self, clear_first: bool = True):
```
ZLLentz, line 151: I like this a lot

happi/client.py
```diff
@@ -526,6 +565,7 @@ def search_range(self, key, start, end=None, **kwargs):
                                               beamline='HXR')
         """
 
+        self._maybe_clear_cache()
```
ZLLentz, line 568: Is there ever a situation where we may need to clear the cache at the end of a search too?

ZLLentz:
CI fails on flake8 of docs/release_notes.py
I guess previously the flake8 must have been ignoring this whole folder

happi/client.py
klauer: I think because the user intent is specified _prior_ to making a request that could hit the backend, the answer is "probably no". I can't think of a scenario within the given API where that could be the case, at least.

tangkong:
Oh also pre-release notes?

ZLLentz (approved):

Issue #124, opened 2020-03-20
Deprecate "device" verbiage for classes from Happi


ZLLentz:
Use `item` instead of `device` to clarify which things are actual functioning devices and which things are descriptions of functioning devices

Todo:
- [x] Deprecate `Device` class, replace with `OphydItem` and `LCLSItem`
- [x] Deprecate `device` module, replace with `item`
- [x] Deprecate methods with `device` in the name except where they return a user's instance, replace with `item`
- [x] Adjust comments, docstrings, and internal code to clarify types
- [x] Clean up the docs in general
- [x] In deprecated classes, ensure documentation refers to replacement class (per @ZryletTC https://github.com/pcdshub/happi/pull/193#discussion_r552331896)

Related:
- [x] Move `LCLSItem` and LCLS's containers into `pcdsdevices`

ZLLentz:
Started in #122

klauer:
- [x] `SearchResult.item`, `Client.all_items` and deprecate `SearchResult.device`, `Client.all_devices`

ZLLentz:
I'm currently getting a headache trying to read through the Client docstrings- it uses device, instantiated device, item, etc. seemingly interchangeably.

klauer:
It's slowly improving but still bad, for sure. Wish we had gotten rid of 'device' longer ago.

PR #266, opened 2022-06-17
REF/CLN: further deprecate usage of 'device' and replace with 'item'

klauer:
## Description
* This is an annoying PR that should be double-checked carefully (sorry)
* Deprecates `*device*` in the client and replaces with `*item*`.

## Motivation and Context
* Addresses https://github.com/pcdshub/happi/issues/124
* It'd be better if this were done sooner rather than later (we suggested this in 2020 and lamented about it in 2021 and here we are in 2022...)

I was considering doing this as part of the speedup PR of #265 but quickly realized how big it'd be, so I pulled it out to a separate branch and rebased after the PR was merged

## How Has This Been Tested?
Test suite.

Issue text and PR text.

## Screenshots

New docs render well enough and include more client-accessing information:
image


ZLLentz:
excellent branch name

happi/backends/json_db.py
```diff
@@ -75,8 +75,8 @@ def _load_or_initialize(self):
         return self._load_cache
 
     @property
-    def all_devices(self):
-        """All of the devices in the database."""
+    def all_items(self):
```
ZLLentz, line 78: From first look, it seems like the approach here is to rename the backend methods and deprecate the client methods- is that correct? I think it's a good approach.

happi/backends/mongo_db.py
```diff
@@ -64,8 +64,8 @@ def __init__(self, host=None, user=None,
                                 'port specified at startup')
```
ZLLentz, line 64: Unrelated to this line: you missed updates in happi/backends/core.py that implicitly set the backend interface

happi/backends/json_db.py
klauer: Yes, that was the intent, at least. Hopefully it's consistent here.

happi/backends/mongo_db.py
klauer: Good catch - 0616a63

I expect this won't be the last you find

happi/item.py
```diff
@@ -59,8 +59,7 @@ class EntryInfo:
     ------
```
ZLLentz, line 59: Is it time to remove `happi/device.py`?

happi/backends/json_db.py
ZLLentz: Reading through very carefully, I think it is

happi/item.py
```diff
@@ -59,8 +59,7 @@ class EntryInfo:
...
     .. code::
 
-        class MyDevice(Device):
-
+        class MyItem(HappiItem):
             my_field = EntryInfo('My generated field')
             number   = EntryInfo('Device number', enforce=int, default=0,
                                  enforce_doc='This must be a number')
```
ZLLentz, line 65: Unrelated to this file: `loader.py` has some arguments confusingly named `device` that are typed as `HappiItem`

happi/tests/test_device.py
```diff
@@ -5,23 +5,22 @@
```
ZLLentz, line 5: Can we rename this file? test_device -> test_item

happi/item.py
klauer: I think so, but I think we should do pcdsdevices first (it relies on `happi.Device` frustratingly)

ZLLentz:
I've finished scouring through the branch for lingering device references- I think the rest are correct but of course I could have missed some too. I'll review the diff now.

happi/tests/test_device.py
klauer: [0f70bf4](https://github.com/pcdshub/happi/pull/266/commits/0f70bf41e7e8c62f1a23db23b4b523d253f18ca6)

happi/item.py
klauer: I missed that entire file, so it seems. A big commit here that needs a careful eye: [0c813f3](https://github.com/pcdshub/happi/pull/266/commits/0c813f3585b3dc11657f6ca1afaf716478daa1c4)

ZLLentz (approved):
Most of this looks fully correct to me

happi/item.py
ZLLentz: oops

ZLLentz: I'll give this a go

docs/source/client.rst
```diff
@@ -141,9 +177,9 @@ easy by way of :meth:`.Client.search_range`. For example:
 
 .. ipython:: python
 
-    client.search_range('z', start=314.4, end=348.6, type='Device')
+    client.search_range('z', start=314.4, end=348.6, type='HappiItem')
```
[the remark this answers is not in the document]
klauer: I... think it is updated? GitHub isn't showing this in the diff somehow?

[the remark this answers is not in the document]
tangkong: `client.search_range('z', start=734.0, stop=735.0, device_class='pcdsdevices.attenuator.FEESolidAttenuator')` --> `at2l0` ?

happi/cli.py
```diff
@@ -321,41 +321,41 @@ def edit(ctx, name: str, edits: List[str]):
...
         logger.info('Aborting')
 
 
 @happi_cli.command()
-@click.argument('device_names', nargs=-1)
+@click.argument('item_names', nargs=-1)
 @click.pass_context
-def load(ctx, device_names: List[str]):
+def load(ctx, item_names: List[str]):
     """Open IPython terminal with DEVICE_NAMES loaded"""
```
tangkong, line 350:
```suggestion
    """Open IPython terminal with ITEM_NAMES loaded"""
```

docs/source/client.rst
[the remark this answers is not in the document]
klauer: I'm all for AT2L0 additions. 🎉 

The only thing is that we need to have something for the docs that exists in happi. That's because these `ipython` sections are actually executed (see the PR body for a screenshot of what these do)

tangkong (approved):
Looks good to me, as far as I can see through the 'device' afterimages burned into my retinas

docs/source/client.rst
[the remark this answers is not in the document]
tangkong: Maybe I've been rendered blind, but where does the `docs_test.json` come from?

happi/cli.py
klauer: Agreed on the change - but in the broader context of parameter names in docstrings:

I've searched `click` documentation and I can't find a reference to how we're intended to format the docstrings here to refer to a parameter. Does it matter? Is it a convention? I had asked @ZLLentz to change `happi benchmark` to `--duration` instead of `DURATION` because (1) I thought it looked ugly and (2) it wasn't really useful to match up with the actual CLI argument.

tangkong: When I first wrote the docstrings I followed this very rough guidance https://click.palletsprojects.com/en/8.1.x/documentation/#documenting-arguments

Apparently there's some Unix convention that I never verified?

docs/source/client.rst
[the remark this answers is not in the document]
klauer: It's a bit weird, but it gets created by those `ipython` blocks in the documentation.

Your comments made me realize that we're actually giving the wrong parameters to those `HappiItem`s and we don't have a good *official* field to even search with `search_range` (though `z` can still work as an extraneous field)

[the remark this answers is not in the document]
ZLLentz: Ah I don't mind too much what we do- I just think it's weird for the example to filter on type at all when that's not a thing people ever do. I was trying to advocate for removing the type= here. This is separate from the scope of this PR but the diff brought it to my attention.

happi/cli.py
klauer: Hmm, the linked documentation makes me think we should be adding `metavar` to these and using that reference in the docstring:

```python
@click.command(options_metavar='<options>')
@click.option('--count', default=1, help='number of greetings',
              metavar='<int>')
@click.argument('name', metavar='<name>')
def hello(count, name):
    """This script prints hello <name> <int> times."""
    for x in range(count):
        click.echo(f"Hello {name}!")
```

`happi profile` for `--profiler` shows:
```
  -p, --profiler TEXT  Select which profiler to use.
```

klauer: At risk of making too big of a deal out of this in the _entirely wrong PR_ I'm going to just accept your suggestion and we can circle back some other time.

Distractions abound...

tangkong: I like that, maybe for another PR 👍

happi/item.py
klauer: Confirmed that branch works with this happi PR 👍

docs/source/client.rst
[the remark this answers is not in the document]
klauer: I made some tweaks to the docs; hopefully they're better than the existing ones

klauer:
Test suite is failing now due to the migration of `Device` to pcdsdevices (and the fact that pcdsdevices is not yet tagged with that change: https://github.com/pcdshub/pcdsdevices/pull/1023)

klauer:
PR still pending pcdsdevices conda package for test suite: https://anaconda.org/conda-forge/pcdsdevices/files

klauer:
OK, tests are passing - anything else I missed here before we merge, @tangkong?

tangkong (approved):
Gave it a quick once over and it looks good to me.  Thanks for sifting through all this!

PR #269, opened 2022-06-28
ENH: allow relative paths for json db in happi config

### LINKED ISSUE #188: JSON database backend file (and other config paths) should be relative to the .cfg file
## Expected Behavior
In a configuration like the following:
```
[DEFAULT]
path = db.json
```

`db.json` should be assumed to be in the path of the `.cfg` file, and not in the current working directory.

## Current Behavior

`db.json` is assumed to be in the current working directory.
As-is, we hard-code this path: `path = /reg/g/pcds/pyps/apps/hutch-python/device_config/db.json` to work around the issue.
This makes using the database where `/reg/g/pcds` is inaccessible a pain.

## Possible Solution
Fix paths prior to sending them to the backend.

JJL772:
Paths to the json db can now be relative to the happi config, unless they're already absolute.

Resolves #188 

## Description

This allows paths to the json db to be relative to the happi config.

For the `device_config` repo, its `happi.cfg` can look like this, instead of containing an absolute path:
```
[DEFAULT]
path=db.json
```

Other relative paths will work fine too, ie `../device_config/db.json`m `./db.json`. Absolute paths are not transformed at all by the json backend when looking for the db, so `/reg/g/pcds/happi.cfg` will still mean `/reg/g/pcds/happi.cfg`. 

The solution I came up with for this isn't as clean as I'd like. Since each backend accepts arbitrary args from the happi config, there's no good way to process paths before sending them to the backend. Ultimately, it's up to the backend implementation to process any paths such that they're correct. Luckily, the only backend that relies on paths from the config is the json backend.

## Motivation and Context

Our happi configs all have absolute paths in them, which makes them annoying to use 
outside of our LCLS environment. For example, I needed to modify the happi.cfg in our `device_config` repo so I could play around with the LCLS device db on my local machine.

## How Has This Been Tested?

One test was modified to account for the new handling, and another test was added to specifically test configs with absolute paths for the json db.

Relevant comments have been modified or created in the code changed by this PR. I didn't see anything in the happi documentation that would need to be changed.

### DIFF
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -7,10 +7,11 @@ import math
 import os
 import os.path
 import re
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 import simplejson as json
 
+from .. import utils
 from ..errors import DuplicateError, SearchError
 from .core import _Backend
 
@@ -49,11 +50,18 @@ class JSONBackend(_Backend):
         Path to JSON file.
     initialze : bool, optional
         Initialize a new empty JSON file to begin filling.
+    cfg_path : str, optional
+        Path to the happi config.
     """
 
-    def __init__(self, path, initialize=False):
+    def __init__(self, path: str, initialize: bool = False, cfg_path: Optional[str] = None):
         self._load_cache = None
-        self.path = os.path.expanduser(path)
+        # Determine the cfg dir and build path to json db based on that unless we're initted w/o a config
+        if cfg_path is not None:
+            cfg_dir = os.path.dirname(cfg_path)
+            self.path = utils.build_abs_path(cfg_dir, path)
+        else:
+            self.path = path
         # Create a new JSON file if requested
         if initialize:
             self.initialize()
--- a/happi/client.py
+++ b/happi/client.py
@@ -849,7 +849,7 @@ class Client(collections.abc.Mapping):
 
         # Create our database with provided kwargs
         try:
-            database = backend(**db_kwargs)
+            database = backend(**db_kwargs, cfg_path=cfg)
             return cls(database=database)
         except Exception as ex:
             raise RuntimeError(
--- a/happi/tests/test_client.py
+++ b/happi/tests/test_client.py
@@ -44,6 +44,29 @@ path=db.json
     os.environ["XDG_CONFIG_HOME"] = xdg_cfg
 
 
+@pytest.fixture(scope='function')
+def happi_cfg_abs(xdg_config_home):
+    # Store current happi config
+    xdg_cfg = os.environ.get("XDG_CONFIG_HOME", '')
+    happi_cfg = os.environ.get("HAPPI_CFG", '')
+
+    # Setup environment variables
+    os.environ['XDG_CONFIG_HOME'] = str(xdg_config_home)
+    os.environ['HAPPI_CFG'] = ''
+
+    # Write file
+    happi_cfg = xdg_config_home / 'happi.cfg'
+    happi_cfg.write_text("""\
+[DEFAULT]
+backend=json
+path=/var/run/db.json
+""")
+    yield str(happi_cfg)
+    # Restore environment variables
+    os.environ["HAPPI_CFG"] = str(happi_cfg)
+    os.environ["XDG_CONFIG_HOME"] = xdg_cfg
+
+
 def test_find_document(happi_client: Client, item_info: Dict[str, Any]):
     doc = happi_client.find_document(**item_info)
     assert doc.pop('prefix') == item_info['prefix']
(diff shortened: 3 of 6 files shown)


happi/backends/json_db.py
```diff
@@ -49,11 +50,18 @@ class JSONBackend(_Backend):
...
     """
 
-    def __init__(self, path, initialize=False):
+    def __init__(self, path: str, initialize=False, cfg_path=None):
         self._load_cache = None
-        self.path = os.path.expanduser(path)
+        # Determine the cfg dir and build path to json db based on that unless we're initted w/o a config
+        if cfg_path is not None:
+            cfg_dir = os.path.dirname(cfg_path)
+            self.path = os.path.expanduser(utils.build_abs_path(cfg_dir, path))
```
ZLLentz, line 62: Is it possible for `build_abs_path` to return a path with unexpanded user directories?

happi/client.py
```diff
@@ -849,7 +849,7 @@ def from_config(cls, cfg=None):
 
         # Create our database with provided kwargs
         try:
-            database = backend(**db_kwargs)
+            database = backend(**db_kwargs, cfg_path=cfg)
```
ZLLentz, line 852: Reviewers note: this is not an API break since there is not yet a supported way to add and use external backends to happi, though this has been a planned feature. (There are unsupported ways to do this)

happi/tests/test_client.py
```diff
@@ -299,10 +322,21 @@ def test_find_cfg(happi_cfg: str):
 
 
 def test_from_cfg(happi_cfg: str):
+    # happi_cfg modifies environment variables to make config discoverable
+    client = Client.from_config()
+    # Internal db path should be constructed relative to the happi cfg dir
+    expected_db = os.path.abspath(os.path.dirname(happi_cfg) + '/db.json')
```
ZLLentz, line 328: For completeness we should use `os.path.join` instead of assuming unix-like paths
```suggestion
    expected_db = os.path.abspath(os.path.join(os.path.dirname(happi_cfg), 'db.json'))
```

happi/utils.py
```diff
@@ -118,3 +119,21 @@ def warn_and_call(*args, **kwargs):
         return warn_and_call
 
     return wrapper
+
+
+def build_abs_path(basedir: str, path: str) -> str:
```
ZLLentz, line 124: I like this utils function :+1:, its definitely on the long long list of "why isn't this in the standard library"

happi/utils.py
```diff
@@ -118,3 +119,21 @@ def warn_and_call(*args, **kwargs):
...
+    Parameters
+    ----------
+    basedir : str
+        If path is not absolute already, build an abspath
+        with path starting here.
+    path : str
+        The path to convert to absolute.
+    """
+    if not os.path.isabs(path):
+        return os.path.abspath(basedir + '/' + path)
```
ZLLentz, line 138: suggest join again for platform compatibility
```suggestion
        return os.path.abspath(os.path.join(basedir, path))
```

ZLLentz (commented):
This issue has been bugging me as well during my offline testing.

Overall this PR is a good implementation. The logic in your PR text is sound. I think having a reference to the config path as an expected backend kwarg is reasonable and I can't think of any way to make this less messy conceptually- alternates like making the backend read and parse the config file itself sound awful. Handing a reference to the file path itself (the only piece of info that isn't in the file) is a good compromise.

If we update the happi configs I'm slightly worried about compatibility with old versions of happi that don't use this feature... I'm not entirely sure what to do about this. But that's no reason to not fix this for future use.

I left a few small suggestions and questions.

JJL772:
> If we update the happi configs I'm slightly worried about compatibility with old versions of happi that don't use this feature... I'm not entirely sure what to do about this. But that's no reason to not fix this for future use.

Maybe we could update configs in a year or so? At that point we shouldn't be running an old version of happi (hopefully :smile:)

happi/utils.py
tangkong: I've been using pathlib for a while and they support it: [p.resolve()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve)

happi/backends/json_db.py
```diff
@@ -49,11 +50,18 @@ class JSONBackend(_Backend):
         Path to JSON file.
     initialze : bool, optional
         Initialize a new empty JSON file to begin filling.
+    cfg_path : str, optional
+        Path to the happi config.
     """
 
-    def __init__(self, path, initialize=False):
+    def __init__(self, path: str, initialize=False, cfg_path=None):
```
tangkong, line 57: While we're at it, do you think you could type annotate the rest of these signatures?  (at least the ones you're modifying)  It's something we're trying to be better about, and every little bit helps.

happi/utils.py
ZLLentz: I don't see any function in the standard library for prepending a base directory if the tail directory is not absolute though

tangkong (commented):
I like the implementation as well.  I don't have much to add on top of what Zach mentioned in his comments.  Since this is backwards compatible I have no qualms about merging

ZLLentz:
> Maybe we could update configs in a year or so? At that point we shouldn't be running an old version of happi (hopefully 😄)

Maybe a bit sooner during the next downtime we could switch over and see if we broke anything

happi/utils.py
tangkong: Ah I see what you mean.  That is inconvenient

tangkong (approved):
LGTM.  Nice work!

PR #270, opened 2022-06-30
REF/FIX: more lazily load the happi container registry

klauer:
## Description
* Load the happi container registry on the first access and not on import
* Add documentation about the registry
* Add type hints for the registry
* Add a companion `__setitem__` for the ability to externally add items to the container registry.
    * That is, `registry[name] = ContainerClass`

## Motivation and Context
* Fixes scenario where a module including a happi container also attempts to import from happi (as @tangkong found)

## How Has This Been Tested?
* We don't really have a test suite for the registry
* It's tested implicitly through the client/CLI tests

This PR text.

## Screenshots

image

### DIFF
--- a/happi/containers.py
+++ b/happi/containers.py
@@ -50,6 +50,14 @@ class HappiRegistry:
                 ],
             },
         )
+
+    Containers may also be added manually to the registry by way of:
+
+    .. code::
+
+        import happi
+
+        happi.containers.registry[item_name] = ContainerClass
     """
     #: Has __init__ been called on the registry?
     __initialized: bool
@@ -136,9 +144,13 @@ class HappiRegistry:
             return ".".join([entry_name, module, klass.__name__])
 
         key = make_entry_name()
-        if key in self._registry:
+
+        class_in_registry = self._registry.get(key, None)
+        if class_in_registry is klass:
+            return
+        if class_in_registry is not None:
             raise RuntimeError(f"Duplicated entry found for key: {key} "
-                               f"and class: {klass}")
+                               f"and class: {klass} {class_in_registry}")
         if klass in self._reverse_registry:
             dup_key = self._reverse_registry.get(klass)
             raise RuntimeError(f"Duplicated entry found. Keys: {key} "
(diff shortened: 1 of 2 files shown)


tangkong:
This seems to break `happi container-registry` for me.  I'll keep looking into it though

tangkong:
I just arrived at that too.  We can tackle both in a separate PR.  I'll throw up an issue

happi/containers.py
```diff
@@ -30,25 +86,30 @@ def __new__(cls, *args, **kwargs):
...
 
-    def items(self):
+    def items(self) -> Generator[Tuple[str, Type[HappiItem]], None, None]:
+        if not self._loaded:
+            self.load()
         yield from self._registry.items()
 
     def entry_for_class(self, klass):
-        if klass not in self._reverse_registry:
+        if not self._loaded:
```
tangkong, line 108: Switching to this conditional means that we don't try to reload the registry if we can't find a class, right?  That seems like nice functionality to have, though it seems rare to get a new container at runtime.

tangkong (approved):
This indeed fixes the issue I saw.  Another step on the path to complete laziness!

happi/containers.py
klauer: That's right. I think the original implementation isn't quite right, really. The edge cases I can imagine are either not really possible to hit or ones we might not want to support:

1. Someone mutates the `DEFAULT_REGISTRY` in `happi.containers` and hopes it'll be picked up after `__getitem__` falls through
2. Someone monkeypatches entrypoints to pick up a new set of containers (this could be fine in a test suite, but just call `.load()` if you need to do this)
3. One I'm unsure of: can you launch `ipython`, import happi, and in a separate terminal do a `pip install` and see updated entrypoints in the original Python instance? I think the answer is "probably not". And even if so, do we care to support this?

klauer: In the interest of keeping things more or less the same as the master branch - whether or not the external `pip install` stuff makes sense to support - I've added that in and will mark this resolved

ZLLentz:
I only saw the one thing- I think this is a very positive change

happi/containers.py
```diff
@@ -103,8 +182,13 @@ def valid_entry(klass):
             return inspect.isclass(klass) and issubclass(klass, HappiItem) \
                    and not klass.__module__.startswith('happi.')
 
-        self._registry = {k: v for k, v in DEFAULT_REGISTRY.items()}
-        self._reverse_registry = {v: k for k, v in self._registry.items()}
+        self._loaded = True
```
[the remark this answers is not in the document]
klauer: I was trying to consider what would happen if `load()` raised unexpectedly in both scenarios (set early/set late) and never settled the debate in my head.

I'll push it to the end of the method and call it good enough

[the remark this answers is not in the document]
ZLLentz: I was trying to think this through too- I think I decided that by the nature of how this function handles extra loads (it handles them well) it'd be better to mark it at the end so we'd by default try again... I think

PR #273, opened 2022-06-30
REL: v1.14.0

klauer:
v1.14.0
=====

API Changes
-----------
- Added ``happi.Client.create_item`` and deprecated ``happi.Client.create_device``.
- Added ``happi.Client.add_item`` and deprecated ``happi.Client.add_device``.
- Added ``happi.Client.find_item`` and deprecated ``happi.Client.find_device``.
- Added ``happi.Client.all_items`` and deprecated ``happi.Client.all_devices``.
- Added ``happi.Client.remove_item`` and deprecated ``happi.Client.remove_device``.
- Deprecated ``happi.SearchResult.device`` and above deprecated items are now scheduled for removal in the next major happi release.
- Internal backend API ``all_devices`` has been changed to ``all_items``.
- ~Removed previously-deprecated ``happi.Device`` container.~ (see below)
- Added ``happi.Client.retain_cache_context`` for clients that desire to control when reloading the database from a happi backend happens.
- Backend implementations may now optionally support a caching mechanism with ``clear_cache`` being called externally by the client when desirable.
- The happi container registry now supports adding new container classes manually by way of ``happi.containers.registry["ContainerName"] = ContainerClass``.

Features
--------
- Significant performance increase for JSON-backed happi clients.
- Makes ``SearchResult`` hashable
- Uses hashable ``SearchResult`` in happi search cli command

Bugfixes
--------
- Fix a rare race condition related to reading a json device database twice in a command line search command between database updates.
- Issue where happi Client would repeatedly (and unnecessarily) make database backend calls has been fixed.
- Allow int search values to match their float counterparts
- The happi container registry is loaded at first use and not on import.  This can result in increased performance where the happi database is not used. It also fixes a scenario in which a module that defines a happi container attempts to import certain classes from happi.

Maintenance
-----------
- Old terminology for ``HappiItem`` instances has been scrubbed and clarified in documentation.
- Test suite and documentation has been updated to reflect trajectory of deprecated methods and naming.
- Added relaxed flake8 configuration.
- Remove happi.device.Device from tests to avoid deprecation warnings
- Add type annotations to test suite
- Clean up fixture usage and separate ``three_valves`` fixture into ``three_valves`` and ``client_with_three_valves``
- add pre-release notes scripts
- More documentation about the happi container registry was added.

Contributors
------------
- klauer
- tangkong
- zllentz


docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W
```
klauer, line 5: See commit note:
```
    Sphinx build option making warnings into errors was removed.
    This is due to the failed glob in a new release:

      upcoming_changes.rst:4: WARNING: toctree glob pattern
      'upcoming_release_notes/[0-9]*' didn't match any documents


```

docs/source/releases.rst
```diff
@@ -1,6 +1,68 @@
-=================
- Release History
-=================
```
klauer, line 3: Title was adjusted to use the pcdsdevices style, as the pre-release document aggregation script butchered the title when I ran it.

tangkong (approved):
LGTM.  🎉

docs/source/releases.rst
```diff
@@ -1,6 +1,68 @@
...
+Release History
+###############
+
+
+v1.14.0 (2022-06-30)
```
ZLLentz, line 5: Does this need to hit v2.0.0 because of the removal of the `Device` container?

ZLLentz: I'd be fine with tagging as-is or doing (1), just want to make sure this isn't a slip up

klauer: For us, I'm hoping to side-step this entirely as removal of a container or differing definitions in development/production environments makes this tough for us to fully support: https://github.com/pcdshub/device_config/pull/29

For any downstream users (?) that might have `Device`-based items, I think (1) would be appropriate to follow with semantic versioning deprecation. I'll push this back to draft until we get that resolved.

klauer:
I think everything is in place for this finally. Anything else I might be missing?

tangkong:
Looks right to me.  👍

PR #282, opened 2022-07-15
ENH: allow filtering of kwargs that have a value matching the default

### LINKED ISSUE #281: Containers need to be more flexible
## Current Behavior
Creation of new containers that build off of old containers is awkward
- Simple inheritance works, but you have to copy class attributes to modify them
- multiple inheritance does not easily combine modifications to a single attribute (if both parents modify kwargs, only one will be seen)

EntryInfo for kwargs can be rigid
- If a kwarg is provided and filled with via jinja, it will always fill that kwarg even if the data does not exist
  - This manifested while implementing `LightpathItem`.  
  - If there is no EntryInfo for kwarg data, the data type (list) may not be preserved
  - init kwargs will always be strings unless an entry exists in the container ([src](https://github.com/pcdshub/happi/blob/f348c1ae9dda82245a631af802d56a766b2e0228/happi/loader.py#L61-L72))

## Possible Solution(s)
Thoughts on inheritance/mixin behavior
- thinking in progress...

Thoughts on kwarg flexibility
- it would be nice if we could have an optional kwarg EntryInfo, such that the kwarg is only submitted if there is valid data for it (non-static kwarg config)
  - could involve a more complicated `EntryInfo`, at the expense of this behavior being hard to infer from the json database
- Could `literal_eval` kwarg arguments in LightpathMixin __init__ to sidestep the issue
- Could annotate the class variables and cast them on __init__ (via `inspect.signature(cls)`)
- Jinja templates / filters?

## Context
This discussion was inspired by my efforts to apply the Lightpath interface to our database.  
In order to store the correct data in the database, we need `input_branches` and `output_branches` to be submitted to the device as kwargs from the database.

If we don't have an EntryInfo for this data specifying the data type (list), the json data will end up as a single string `"['L0', 'L1']"`

To apply an EntryInfo to all the different containers in `pcdsdevices`, we would have to either mixin (which fails), or define a bunch of additional classes (which is pain)

Further complicating the issue is that currently there are cases where a single container corresponds to multiple device classes, some needing the lightpath interface and some not.  
- if we use lightpath container --> default kwargs always has branches --> some devices get extra kwargs --> error
- If we don't use lightpath container --> data that fills the kwargs jinja template gets interpreted as a str instead of a list --> error when trying to read branch data

<details>
  <summary>Example, where we fill a `LCLSItem` container with data from `LightpathItem`.</summary>

```python

In [6]: sr = res[0]

In [7]: sr.metadata
Out[7]: 
{'name': 'sl1k2',
 'device_class': 'lightpath.tests.conftest.Stopper',
 'args': ['{{prefix}}'],
 'kwargs': {'input_branches': '{{input_branches}}',
  'output_branches': '{{output_branches}}',
  'z': '{{z}}',
  'name': '{{name}}'},
 'active': True,
 'documentation': None,
 'prefix': 'SL1K2:L2SI',
 'beamline': None,
 'location_group': None,
 'functional_group': None,
 'z': 778.833,
 'stand': None,
 'lightpath': True,
 'ioc_engineer': None,
 'ioc_location': None,
 'ioc_hutch': None,
 'ioc_release': None,
 'ioc_arch': None,
 'ioc_name': None,
 'ioc_type': None,
 '_id': 'sl1k2',
 'creation': 'Wed Jun 22 09:53:23 2022',
 'input_branches': ['K2'],
 'last_edit': 'Wed Jun 22 09:53:23 2022',
 'output_branches': ['K2'],
 'type': 'pcdsdevices.happi.containers.LCLSItem'}

In [8]: sr.get()
Can not enforce type to match attribute input_branches as it does not exist on the container.
Can not enforce type to match attribute output_branches as it does not exist on the container.
Out[8]: Stopper(prefix='SL1K2:L2SI', name='sl1k2', read_attrs=['current_state', 'current_transmission', 'current_destination', 'lightpath_summary'], configuration_attrs=[])

In [9]: dev = sr.get()

In [10]: dev.output_branches
Out[10]: "['K2']"
```
</details>

A slew of discussions happened on slack, and I've done my best to transfer them here

## Your Environment
pcds-5.4.1

tangkong:
## Description
- Adds the kwarg `include_default_as_kwarg` to `EntryInfo`
- Adds a clause in `from_container` to omit `kwargs` if their value is equal to the default stored on the corresponding `EntryInfo`, triggered by the `item._info_attrs['kwargs'].include_default_as_kwarg` flag.
  - Gives individual `EntryInfo`'s the ability to decide whether or not to `include_default_as_kwarg` in the case where `item._info_attrs['kwargs'].include_default_as_kwarg = True`
- adds tests for this feature and a fixture that covers jinja template code paths

What is the behavior now?  Considering the case where we are attmpting to load a device from the following document:
```yaml
{
    'name': 'alias',
    ...
    'kwargs': {
        'hi': 'oh hello',
        'blank': '{{blank}}'
        },
    'blank': 'default_blank'
}
```
- `'hi' = 'oh hello'` will always be passed through as a kwarg
- `blank` 's value will depend on the container
  - if `blank` has no corresponding `EntryInfo` --> `blank = 'default_blank'`
    - if include_default_as_kwarg == True: `kwargs = {'hi': 'oh hello', 'blank': 'default_blank'}`
    - if include_default_as_kwarg == False: `kwargs = {'hi': 'oh hello', 'blank': 'default_blank'}` (no effect)
  - if `blank` has an `EntryInfo` and its default is 'default_blank' --> 
    - if include_default_as_kwarg == True: `kwargs = {'hi': 'oh hello', 'blank' = 'default_blank'}`
    - if include_default_as_kwarg == False: `kwargs = {'hi': 'oh hello'}`

- All `EntryInfo`'s have this setting.  You can change this setting either on the kwargs EntryInfo, or a non-kwargs EntryInfo
  - If the kwargs EntryInfo sets `include_default_as_kwarg = True` (default)
    - Consult the corresponding `EntryInfo`'s setting for each kwarg for whether to include the default or not
  - if the kwargs EntryInfo sets `include_default_as_kwarg  = False`
    - Ignore the corresponding `EntryInfo` setting.  Ignore any kwarg whose value matches the EntryInfo default

## Motivation and Context
Closes #281 
I wanted a way to ignore a kwarg if it isn't filled with relevant information.  

Is this the behavior we want?
- The default behavior is unchanged, and the flag makes it possible to opt-in to this new behavior
- I would argue that if a piece of information lacks an `EntryInfo`, it is metadata, and should be carried through
- if there is an EntryInfo and it is being passed as a kwarg, we have to leave it to the user to decide whether they actually want to pass `None` as a kwarg

Why add this flag to all `EntryInfo` instances when only `kwargs` cares?
- I felt that making a separate subclass of `EntryInfo` just for kwargs would cause more confusion than it's worth.  Though I could be convinced othewise

## How Has This Been Tested?
interactively, and a test has been added to the test suite. 

comments, this PR

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/282-enh_filter_default_kwargs.rst
+++ b/docs/source/upcoming_release_notes/282-enh_filter_default_kwargs.rst
@@ -1,0 +1,25 @@
+282 enh_filter_default_kwargs
+#############################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- adds flag ``include_default_as_kwarg``.  If False on a kwargs EntryInfo,
+  any values that match the default on the corresponding EntryInfo will be
+  omitted from the kwarg dictionary.
+  The default value is True, retaining the original behavior.
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tangkong
--- a/docs/source/upcoming_release_notes/282-enh_filter_default_kwargs.rst
+++ b/docs/source/upcoming_release_notes/282-enh_filter_default_kwargs.rst
@@ -7,9 +7,11 @@ API Changes
 
 Features
 --------
-- adds flag ``include_default_as_kwarg``.  If False on a kwargs EntryInfo,
-  any values that match the default on the corresponding EntryInfo will be
-  omitted from the kwarg dictionary.
+- Add `EntryInfo` keyword argument ``include_default_as_kwarg``.  If set to ``False``,
+  any keys that are included in an item's ``kwargs`` that match the default of their
+  corresponding `EntryInfo` will be omitted from the keyword arguments passed to
+  ``device_class`` when instantiating (loading) the item as in ``happi.loader.load_device`` or
+  ``SearchResult.get()``.
   The default value is True, retaining the original behavior.
 
 Bugfixes
--- a/docs/source/upcoming_release_notes/282-enh_filter_default_kwargs.rst
+++ b/docs/source/upcoming_release_notes/282-enh_filter_default_kwargs.rst
@@ -7,12 +7,17 @@ API Changes
 
 Features
 --------
-- Add `EntryInfo` keyword argument ``include_default_as_kwarg``.  If set to ``False``,
+- Add ``EntryInfo`` keyword argument ``include_default_as_kwarg``.  If set to ``False``,
   any keys that are included in an item's ``kwargs`` that match the default of their
-  corresponding `EntryInfo` will be omitted from the keyword arguments passed to
+  corresponding ``EntryInfo`` will be omitted from the keyword arguments passed to
   ``device_class`` when instantiating (loading) the item as in ``happi.loader.load_device`` or
   ``SearchResult.get()``.
-  The default value is True, retaining the original behavior.
+  If the ``kwargs`` EntryInfo sets ``include_default_as_kwarg = True``,
+  the setting on the corresponding ``EntryInfo`` will be used to decide
+  whether or not to omit a keyword argument.
+  If the ``kwargs`` EntryInfo sets ``include_default_as_kwarg = False``,
+  the setting on corresponding ``EntryInfo`` will be ignored.
+  The default value is True on all EntryInfo instances, retaining the original behavior.
 
 Bugfixes
 --------
--- a/happi/item.py
+++ b/happi/item.py
@@ -81,7 +81,7 @@ class EntryInfo:
         enforce: Optional[Any] = None,
         default: Optional[Any] = None,
         enforce_doc: Optional[str] = None,
-        include_default_as_kwarg: Optional[bool] = True
+        include_default_as_kwarg: bool = True
     ):
         self.key = None  # Set later by parent class
         self.doc = doc
--- a/happi/item.py
+++ b/happi/item.py
@@ -50,7 +50,7 @@ class EntryInfo:
         if the entered information does not follow the ``enforce`` type
     include_default_as_kwarg : bool, optional
         Defaults to True.  If a kwargs EntryInfo sets this to False, all kwargs
-        will be compared to their corresponding Entries in the item. and
+        will be compared to their corresponding Entries in the item and
         omitted from the kwargs dictionary if their value matches the Entry's
         default.
         This can also be set on an individual Entry basis.  The setting on
--- a/happi/item.py
+++ b/happi/item.py
@@ -3,6 +3,7 @@ import copy
 import logging
 import sys
 from collections import OrderedDict
+from typing import Any, Optional
 
 from prettytable import PrettyTable
 
@@ -65,13 +66,20 @@ class EntryInfo:
                                  enforce_doc='This must be a number')
     """
 
-    def __init__(self, doc=None, optional=True, enforce=None,
-                 default=None, enforce_doc=None, filter_none=False):
+    def __init__(
+        self,
+        doc: Optional[str] = None,
+        optional: Optional[bool] = True,
+        enforce: Optional[Any] = None,
+        default: Optional[Any] = None,
+        enforce_doc: Optional[str] = None,
+        include_default_as_kwarg: Optional[bool] = True
+    ):
         self.key = None  # Set later by parent class
         self.doc = doc
         self.enforce = enforce
         self.optional = optional
-        self.filter_none = filter_none
+        self.include_default_as_kwarg = include_default_as_kwarg
         self.enforce_doc = str(enforce_doc or '')
 
         # Explicitly set default to None b/c this is how we ensure mandatory
--- a/happi/item.py
+++ b/happi/item.py
@@ -48,6 +48,14 @@ class EntryInfo:
     enforce_doc : str, optional
         A human-readable explanation of the enforce field.  Will be printed
         if the entered information does not follow the ``enforce`` type
+    include_default_as_kwarg : bool, optional
+        Defaults to True.  If a kwargs EntryInfo sets this to False, all kwargs
+        will be compared to their corresponding Entries in the item. and
+        omitted from the kwargs dictionary if their value matches the Entry's
+        default.
+        This can also be set on an individual Entry basis.  The setting on
+        an individual entry will only be taken into consideration if the
+        kwarg EntryInfo has this set to True (default).
 
     Raises
     ------
--- a/happi/item.py
+++ b/happi/item.py
@@ -66,11 +66,12 @@ class EntryInfo:
     """
 
     def __init__(self, doc=None, optional=True, enforce=None,
-                 default=None, enforce_doc=None):
+                 default=None, enforce_doc=None, filter_none=False):
         self.key = None  # Set later by parent class
         self.doc = doc
         self.enforce = enforce
         self.optional = optional
+        self.filter_none = filter_none
         self.enforce_doc = str(enforce_doc or '')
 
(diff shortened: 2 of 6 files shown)


tangkong:
This is my moderately hacky solution to the kwargs problem.  It was the least invasive / api breaking way I could think of, but I'm open to critiques.

klauer:
> if filter_none == False: kwargs = {'hi': 'oh hello', 'blank' = NoneType}

From the PR text - is this true? `blank=type(None)`?

I see what you meant now about filtering out `None`. I have mixed feelings about the entire `kwarg` mechanism overall, but this feature seems not unreasonable when made opt-in.

The fact that `filter_none` is specific to how kwargs get populated makes me think the name could at least be better.
Maybe: `skip_kwarg_if_none` / `skip_none_as_kwarg` / `include_none_as_kwarg`

Slightly different thought: `include_default_as_kwarg`
Perhaps `None` shouldn't be special for skipping, but rather the container default EntryInfo value?

tangkong:
> From the PR text - is this true? `blank=type(None)`?

Sorry, not the type of `None`, but `None` itself.  I was concerned with differentiating between the string and `None` itself,  but ended up being more confusing along the way.  (I've edited the pr text to match)

> Slightly different thought: `include_default_as_kwarg` Perhaps `None` shouldn't be special for skipping, but rather the container default EntryInfo value?

As in we skip the kwarg if there template is filled with the default value stored in the EntryInfo?  What if said default is meaningful?  (and I suppose different from the default value on the device). 

I'm happy to switch names around etc

klauer:
> Sorry, not the type of `None`, but `None` itself. I was concerned with differentiating between the string and `None` itself, but ended up being more confusing along the way. (I've edited the pr text to match)

Gotcha, OK. It would have been even worse were it `NoneType`.

> As in we skip the kwarg if there template is filled with the default value stored in the EntryInfo? What if said default is meaningful? (and I suppose different from the default value on the device).

If it's meaningful, you wouldn't be setting `include_default_as_kwarg=True`, presumably?

tangkong:
Oh you're right.  I can get on board with `include_default_as_kwarg`, which boils down to the same as I've implemented when default=None.  It is less arbitrary than `None` I suppose

klauer:
I think this looks good and would be OK moving forward unless others feel strongly otherwise

ZLLentz:
I meant to give this a review last week but ran out of time with my late meeting- I'll look now

ZLLentz:
> I felt that making a separate subclass of EntryInfo just for kwargs would cause more confusion than it's worth. Though I could be convinced othewise

I think it might be worth revisiting this later, but that doesn't mean we shouldn't include the option described by this PR first

tangkong:
> I felt that making a separate subclass of EntryInfo just for kwargs would cause more confusion than it's worth. Though I could be convinced otherwise

If there was more functionality than just this filtering behavior, I think it could be justified.  The args/kwargs EntryInfos are already treated differently, after all

docs/source/upcoming_release_notes/282-enh_filter_default_kwargs.rst
```diff
@@ -0,0 +1,25 @@
+282 enh_filter_default_kwargs
+#############################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- adds flag ``include_default_as_kwarg``.  If False on a kwargs EntryInfo,
```
klauer, line 10: Some suggestions - take of it what you will:
```
Add `EntryInfo` keyword argument ``include_default_as_kwarg``.  If set to ``False``,
any keys that are included in an item's ``kwargs`` that match the default of their
corresponding `EntryInfo` will be omitted from the keyword arguments passed to
``device_class`` when instantiating (loading) the item as in ``happi.loader.load_device`` or
``SearchResult.get()``.
The default value is True, retaining the original behavior.
```

happi/loader.py
```diff
@@ -177,6 +177,14 @@ def create_arg(arg):
     args = [create_arg(arg) for arg in item.args]
     kwargs = dict((key, create_arg(val))
                   for key, val in item.kwargs.items())
+    # maybe filter out null kwargs
+    if not item._info_attrs['kwargs'].include_default_as_kwarg:
+        new_kwargs = {}
+        for k, v in kwargs.items():
+            einfo = item._info_attrs.get(k, False)
+            if not einfo or not einfo.default == v:
+                new_kwargs[k] = v
```
ZLLentz, line 186: I always find single-letter variable name loops hard to read. This is maybe a personal preference of mine rather than a complaint here. I probably would have used `entry`, `value` and agonized over whether that was the clearest word choice.

happi/loader.py
```diff
@@ -177,6 +177,14 @@ def create_arg(arg):
     args = [create_arg(arg) for arg in item.args]
     kwargs = dict((key, create_arg(val))
                   for key, val in item.kwargs.items())
+    # maybe filter out null kwargs
+    if not item._info_attrs['kwargs'].include_default_as_kwarg:
```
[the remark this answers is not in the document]
tangkong: The way it's implemented, either all kwargs will be subject to filtering, or none of them will.  This property gets set on the kwargs EntryInfo, rather than individually for each EntryInfo that might get pulled into a kwarg.  

Am I this question right?

happi/loader.py
tangkong: I'm not opposed to more full variable names, I suppose I use these single letter names if I really don't intend for them to be used elsewhere.

happi/loader.py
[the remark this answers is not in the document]
ZLLentz: Got it, I was confused because of the seeming (near) contradiction between the above implementation and the desire to not do something special with the kwargs `EntryInfo`. I realize that you need to redefine `kwargs` anyway in these subclasses so the distinction is pretty minor.

docs/source/upcoming_release_notes/282-enh_filter_default_kwargs.rst
```diff
@@ -0,0 +1,27 @@
...
+- N/A
+
+Features
+--------
+- Add `EntryInfo` keyword argument ``include_default_as_kwarg``.  If set to ``False``,
+  any keys that are included in an item's ``kwargs`` that match the default of their
+  corresponding `EntryInfo` will be omitted from the keyword arguments passed to
+  ``device_class`` when instantiating (loading) the item as in ``happi.loader.load_device`` or
+  ``SearchResult.get()``.
+  The default value is True, retaining the original behavior.
```
ZLLentz, line 15: We should either clarify that this needs to be used on the `kwargs` entry info or allow it to have some associated effect when used on the others. As-written this is not entirely clear about how to invoke this behavior.

happi/item.py
```diff
@@ -65,12 +66,20 @@ class MyItem(HappiItem):
...
-    def __init__(self, doc=None, optional=True, enforce=None,
-                 default=None, enforce_doc=None):
+    def __init__(
+        self,
+        doc: Optional[str] = None,
+        optional: Optional[bool] = True,
+        enforce: Optional[Any] = None,
+        default: Optional[Any] = None,
+        enforce_doc: Optional[str] = None,
+        include_default_as_kwarg: Optional[bool] = True
```
ZLLentz, line 76: I think this deserves a docstring call-out above that makes it clear where/when to use it

ZLLentz (approved):
Happy to merge

I'm having trouble following the loop logic in the happi loader... but I think it's all solid

klauer (approved):

tangkong:
This has proven useful in updating the happi database for Lightpath compatibility, thanks for the comments everyone

PR #284, opened 2022-08-01
MNT: set a max width for item.show_info tables

tangkong:
## Description
Calculates a max width for tables generated by `item.show_info()` based on the current terminal width

## Motivation and Context
After I added a bunch of kwargs, the tables got annoying to view.  

## How Has This Been Tested?
Interactively

This PR.

Before
image

After
image

## Pre-merge checklist

### DIFF
--- a/happi/item.py
+++ b/happi/item.py
@@ -1,6 +1,7 @@
 import collections
 import copy
 import logging
+import os
 import sys
 from collections import OrderedDict
 from typing import Any, Optional
@@ -335,7 +336,6 @@ class HappiItem(_HappiItemBase, collections.abc.Mapping):
         """
 
         pt = PrettyTable(['EntryInfo', 'Value'])
-        pt._max_width = {'Value': 60}
         pt.align = 'r'
         pt.align['EntryInfo'] = 'l'
         pt.align['Value'] = 'l'
@@ -348,6 +348,15 @@ class HappiItem(_HappiItemBase, collections.abc.Mapping):
                               if not key.startswith('_')])
         for key in public_keys:
             pt.add_row([key, show_info[key]])
+
+        # initialize table so we can grab widths
+        _ = pt.get_string()
+
+        # set width of Value column dynamically
+        # set width(Value) = terminal width - width(EntryInfo column)
+        entry_info_width = pt._widths[0] + 10  # account for padding
+        term_width = os.get_terminal_size()[0]
+        pt._max_width = {'Value': max(60, term_width - entry_info_width)}
 
         print(pt, file=handle)
 
--- a/happi/item.py
+++ b/happi/item.py
@@ -335,6 +335,7 @@ class HappiItem(_HappiItemBase, collections.abc.Mapping):
         """
 
         pt = PrettyTable(['EntryInfo', 'Value'])
+        pt._max_width = {'Value': 60}
         pt.align = 'r'
         pt.align['EntryInfo'] = 'l'
         pt.align['Value'] = 'l'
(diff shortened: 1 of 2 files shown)


happi/item.py
```diff
@@ -335,6 +335,7 @@ def show_info(self, handle=sys.stdout):
         """
 
         pt = PrettyTable(['EntryInfo', 'Value'])
+        pt._max_width = {'Value': 60}
```
ZLLentz, line 338: This is the width of the full table?

I think a max width is appropriate and 60 is reasonable, but I wonder if we should also think about getting the user's terminal width to help us set this properly.

ZLLentz: Also: is there some better way to do this than to use a semi-private variable? Just `max_width` maybe?

tangkong: You can set the max width of the whole table via `pt.max_table_width` and the width by column via `pt._max_width`

Getting the user's terminal width would be cool, I didn't think of that.  It looks easy enough to do too.

ZLLentz: It looks like there is also a `max_width` property that reads from `_max_width`- might want to avoid poking at the internals if we don't have to.

tangkong: Also `max_width` will also steal space from the other columns, which I didn't want.  I'll see if I can't refine this more

happi/item.py
```diff
@@ -348,6 +349,15 @@ def show_info(self, handle=sys.stdout):
         for key in public_keys:
             pt.add_row([key, show_info[key]])
 
+        # initialize table so we can grab widths
+        _ = pt.get_string()
+
+        # set width of Value column dynamically
+        # set width(Value) = terminal width - width(EntryInfo column)
+        entry_info_width = pt._widths[0] + 10  # account for padding
+        term_width = os.get_terminal_size()[0]
```
klauer, line 358: I seem to recall that `get_terminal_size` can fail if using a program in non-interactive mode (e.g., piping its output to a file). Can you do a quick check of `happi search * > output.txt` for example?

I'd expect you'll need a try/except around this for that case.

happi/item.py
```diff
@@ -348,6 +349,15 @@ def show_info(self, handle=sys.stdout):
...
+        pt._max_width = {'Value': max(60, term_width - entry_info_width)}
```
klauer, line 359: 80 is the "standard" terminal width, if there is one. Why 60?

tangkong: I'm only modifying the width of the `Value` column, so it was originally 80 - (estimated width of the EntryInfo column)

tangkong:
Ahahaah this breaks all the hard coded tests.  If it isn't the consequences of my own actions

happi/item.py
tangkong: Excellent catch.  Fix on the way

ZLLentz:
Oh no, I forgot about all the explicit output tests... What a pain.

tangkong:
Actually I spoke too soon.  Most of the travis errors are due the non-interactive mode Ken mentioned.  We actually managed to avoid explicit output tests for the search / show_info parts, though there are a bunch of other tests that I'm sure we'll have to break later.

tangkong:
Any other thoughts?

klauer:
One last thought, and would be good to merge in either case:
I think that for piping results to a file, I'd prefer it to not have a maximum width set at all (or a very high one).
While we _should_ be working with the supported `--json` output, if we're lazy and using `tr`/`sed`/`awk` etc, it'd be easier to not have to worry about the wrapping.

tangkong:
sounds good.  I just need a rubber stamp then and I can merge.

klauer (approved):
LGTM 👍

PR #286, opened 2022-09-14
ENH: happi audit

tangkong:
## Description
- Adds a happi audit command that runs checks on devices or subsets thereof.
- Adds a few basic initial checks
  - check_instantiation: can we create the device?
  - check_extra_info: checks for information in the db that is not in the container
  - check_name_match_id: checks that the device's `get_lightpath_state` method works and is reasonable
- Moves some fixtures around
- Touches up cli docstrings to make the more readable in the `happi -h` message
- Sneaks in a fix that restores the old `happi search --json` API ([e5c8255](https://github.com/pcdshub/happi/pull/286/commits/e5c8255c2463e7fce8fb8c5363970401a65d0692))

## Motivation and Context
closes #168 
I was doing similar things to test the lightpath database, so I spent some time cleaning it up and putting it here

**What about and EntryInfo enforcement check?**
Turns out that if information fails to be enforced, we can't even load the SearchResult in the first place.   If this issue exists, it'll pop up during the search function.  ~~I wrote a function and forgot to take it out before committing, but this check might need to be re-scoped~~ This check has been removed

## How Has This Been Tested?
interactively

this PR
<details>
  <summary>Console output examples</summary>

```console
(pcds-5.4.2)roberttk@psbuild-rhel7-01:~/devrepos/happi(enh_audit -)$ happi audit -c inst at2l0
found 1 items to verify
running checks: ['check_instantiation']
# devices failed: 0 / 1
```

``` console
(pcds-5.4.2)roberttk@psbuild-rhel7-01:~/devrepos/happi(enh_audit -)$ happi audit -c _id '*'
found 657 items to verify
running checks: ['check_name_match_id']
+------------+---------------------+------------------------------------+
|    name    |        check        |               error                |
+------------+---------------------+------------------------------------+
| mr1l3_homs | check_name_match_id | id: MR1L3_homs != name: mr1l3_homs |
| mr1l4_homs | check_name_match_id | id: MR1L4_homs != name: mr1l4_homs |
| mr2l3_homs | check_name_match_id | id: MR2L3_homs != name: mr2l3_homs |
+------------+---------------------+------------------------------------+
# devices failed: 3 / 657
```
```console
(pcds-5.4.2)roberttk@psbuild-rhel7-01:~$ happi audit -f ~/test/happi_audit_fns.py -c light -c extra at*l*
found 5 items to verify
running checks: ['check_lightpath_valid', 'check_extra_info']
+------------+-----------------------+------------------------------------------------------------------------+
|    name    |         check         |                                 error                                  |
+------------+-----------------------+------------------------------------------------------------------------+
| at1k4_calc | check_lightpath_valid |                          get_lightpath_state                           |
|            |    check_extra_info   |   Un-enforced metadata found: ['calculator_prefix', 'motor_prefix']    |
|   at1l0    | check_lightpath_valid |            reduce() of empty sequence with no initial value            |
| at2k2_calc | check_lightpath_valid |                          get_lightpath_state                           |
|   at2l0    | check_lightpath_valid |            reduce() of empty sequence with no initial value            |
|            |    check_extra_info   |   Un-enforced metadata found: ['detailed_screen', 'embedded_screen',   |
|            |                       |          'engineering_screen', 'macros', 'parent', 'system']           |
| at2l0_calc | check_lightpath_valid |                          get_lightpath_state                           |
+------------+-----------------------+------------------------------------------------------------------------+
# devices failed: 5 / 5
```

```console
(pcds-5.4.2)roberttk@psbuild-rhel7-01:~/devrepos/happi(enh_audit -)$ happi audit -f ~/test/happi_audit_fns.py -l
+-----------------------+------------------------------------------------------------------+
|          name         | description                                                      |
+-----------------------+------------------------------------------------------------------+
|  check_instantiation  | Check if the device can be instantiated                          |
|                       | Simply attempts to get the device from the happi.SearchResult    |
|                       |                                                                  |
|                       | Subscription callback exceptions are silenced in ophyd, and will |
|                       | not cause this check to fail.                                    |
+-----------------------+------------------------------------------------------------------+
|    check_extra_info   | Check if there is any extra info in the result that does not     |
|                       | match the container.                                             |
|                       |                                                                  |
|                       | Ignores the presence of client-side metadata                     |
|                       | - id                                                             |
|                       | - type                                                           |
|                       | - creation                                                       |
|                       | - last_edit                                                      |
+-----------------------+------------------------------------------------------------------+
|  check_name_match_id  | Check if the item's ``_id`` field matches its ``name``.          |
|                       | This is a convention we have held for a while, making searching  |
|                       | in either direction easier.                                      |
+-----------------------+------------------------------------------------------------------+
| check_lightpath_valid | Check if the desired device is lightpath-valid by:               |
|                       | - verifying device.get_lightpath_state() runs                    |
|                       | - received LightpathState is valid                               |
+-----------------------+------------------------------------------------------------------+
```
</details>

## Pre-merge checklist


klauer:
I'm excited to see this one moving forward again 🎉 

I just ran into a new one that this tool could help with: `_id` and `name` differing. 
https://github.com/pcdshub/device_config/blob/master/db.json#L3-L23
At least this is a topic for discussion: I was of the opinion that `name` and `_id` should be the same thing so that we can look up in either direction easily. If we all agree that's the way forward, such a check could be part of happi audit. In the future we could look to reduce redundancy in the database as `_id` becomes superfluous.

tangkong:
I added the ability to pull functions from external files to the CLI.  I expect it to be a very useful feature, but it involves modifying sys.path and importing from user provided files.  It's less than safe, but I had trouble coming up with a better way to do it.  I added some checks but there's plenty of adversarial ways to get around my validation..

ZLLentz:
This supersedes and closes #168

happi/audit.py
```diff
@@ -0,0 +1,137 @@
...
+
+def check_instantiation(result: SearchResult) -> None:
+    """
+    Check if the device can be instantiated
+    Simply attempts to get the device from the happi.SearchResult
+
+    Subscription callback exceptions are silenced in ophyd, and will
+    not cause this check to fail.
+    """
+    _ = result.get()
```
ZLLentz, line 24: Nitpick, but why not just:
```suggestion
    result.get()
```

happi/cli.py
```diff
@@ -873,6 +884,125 @@ def no_create_channel(*args, **kwargs):
...
+    successful.
+
+    To import additional checks, provide a file with your check function
+    and a list named ``checks`` containing the desired functions.
+    """
+    logger.debug('Starting audit block')
+    from .audit import checks, verify_result
+
+    # if a file is provided, make its functions available
+    if ext_file:
```
ZLLentz, line 928: This is kind of neat but also kind of clunky. I wonder if this is a good use case for an entrypoint, so then you could define the lightpath check in lightpath for example.

happi/cli.py
```diff
@@ -873,6 +884,125 @@ def no_create_channel(*args, **kwargs):
...
+        click.echo('\n')
+        click.echo(' '.join(test_results['name']))
+    else:
+        pt = prettytable.PrettyTable(field_names=['name', 'check', 'error'])
+        for name, success, msg in zip(test_results['name'],
+                                      test_results['success'],
+                                      test_results['msg']):
+            if not success:
+                check_text, error = msg.split(': ', 1)
+                check_name = re.search(r'\((.*)\)', check_text)[1]
```
ZLLentz, line 991: It seems odd to me that you need to regex through your own return value to get this info, I'd imagine this would be simpler to embed in the test_results

ZLLentz (approved):
A few nitpicks but by and large I'm excited to be able to run this. I want to add it to our device_config repo's CI.

happi/cli.py
tangkong: I thought about setting up entrypoints, but though that might also be too clunky, since you have to install a package to get the entrypoints to pick up.  I agree it's sort of ham-fisted, aside from the signature checks it's the part of this I like the least.

Maybe the world where a random user comes up with a check and wants to quickly run it is too fanciful.

happi/audit.py
tangkong: That's a good question, seeing as I do that with the output of the check functions themselves...

happi/cli.py
tangkong: erm.  Yes.  That makes more sense..

happi/audit.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    success, check, msg = True, "", ""
+    # verify checks have the correct signature, as much as is reasonable
+    for check in checks:
+        validate_check_signature(check)
+
+    try:
+        for check in checks:
+            check(result)
+    except Exception as ex:
```
klauer, line 96: Would there be a benefit to showing _all_ checks that failed instead of just the first check?

happi/audit.py
```diff
@@ -0,0 +1,140 @@
...
+    Raises
+    -------
+    RuntimeError
+        if function cannot be used to audit happi SearchResult's
+    """
+
+    if not callable(fn):
+        raise RuntimeError('requested check function is not callable')
+
+    sig = inspect.getfullargspec(fn)
```
klauer, line 127: Though what you have is just fine, I find that `sig.bind()` is usually good enough for these checks:
```python
sig = inspect.signature(fn)
sig.bind(search_result)
```

happi/cli.py
```diff
@@ -873,6 +883,125 @@ def no_create_channel(*args, **kwargs):
...
+    Checks are simple functions that raise exceptions on failure,
+    whether naturally or via assert calls.  These functions take a single
+    happi.SearchResult as an positional argument and returns None if
+    successful.
+
+    To import additional checks, provide a file with your check function
+    and a list named ``checks`` containing the desired functions.
+    """
+    logger.debug('Starting audit block')
+    from .audit import checks, verify_result
```
klauer, line 924: Any reason for the imports inside this function? (Here and importlib below)

klauer (approved):
Awesome

happi/audit.py
tangkong: When I started this I think I was envisioning subsequent checks would be nonsensical if earlier checks failed (checking if methods exist on devices only makes sense if the device can be instantiated).  But with the given checks, I think it makes sense to run all the checks and show all failures.

happi/audit.py
tangkong: Oh that's neat.  I clearly needed to read through more of the `inspect` module

happi/cli.py
tangkong: I figured that since they were only used by the audit command I didn't need to burden the rest of the file with the imports.  But I guess they're lightweight enough that we can be pythonic and move them to the top

klauer: I see, OK. Sometimes people do this because it results in a circular import - which can (sometimes/often) be a red flag.

happi/audit.py
klauer: I think what you have proposed is plenty good for a first pass. This could be a follow-up or just something that we file away in an issue - your call

tangkong: addressed in [c2a4bc9](https://github.com/pcdshub/happi/pull/286/commits/c2a4bc95c34499b12ac90d45740425c731927ec0)

happi/cli.py
```diff
@@ -873,6 +885,151 @@ def no_create_channel(*args, **kwargs):
...
+
+    unique_fails = set(test_results['name'][i]
+                       for i in range(len(test_results['name']))
+                       if not test_results['success'][i])
+    # print outs
+    if names_only:
+        click.echo(' '.join(unique_fails))
+    elif show_json:
+        final_dict = {'audited': len(results),
+                      'falures': len(unique_fails)}
```
klauer, line 993: I just noticed the additional `--json` feature. People and their side-channel requests 🙄 

Looks good! Just one little typo:
```suggestion
                      'failures': len(unique_fails)}
```

tangkong: I used to be good at spelling sigh

tangkong:
Alrighty, in this goes.  Follow up issues sure to follow, but hopefully this finds some use.

PR #287, opened 2022-10-05
ENH: Simultaneous Multi-Backend support

### LINKED ISSUE #24: Support multiple configurations

tangkong:
## Description
- Creates MultiBackend, which itself takes multiple backends.
  - MultiBackend prioritizes backends in the order they are listed in the config, with the DEFAULT config used last

## Motivation and Context
closes #24 ... some things never change
If users have small, hutch-specific configuration changes to the happi database, previously they would have to direct happi to their own version of the database.  This allows users to incorporate a small db file on top of our facility-wide device_config, which acts as a fallback.

### Config file
Currently our config files have only the path to a single db file, with the option to specify a backend and its associated initialization arguments

```
[DEFAULT]
path = /cds/home/r/path/to/db.json
```

This PR allows sections other than `DEFAULT` to specify additional backends.   If more than one backend (section) is detected at config-read-time, a MultiBackend will be initialized with those backends.  `DEFAULT` is given the lowest priority, and the remaining backends will be prioritized in order of appearance.

```
[CONFIG1]
# will take precedence over DEFAULT and CONFIG1
path = /cds/home/r/path/to/db1.json

[DEFAULT]
path = /cds/home/r/path/to/db.json

[CONFIG2]
# will take precedence over DEFAULT, but duplicates present in CONFIG1 will not be seen
path = /cds/home/r/path/to/db2.json
```

## How Has This Been Tested?
A test has been added, interactive testing 

This PR, docstrings, type-hints.  Config file documentation is forthcoming, if this looks good

## Pre-merge checklist

### DIFF
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -173,10 +173,10 @@ class JSONBackend(_Backend):
             except Exception as ex:
                 logger.debug('Comparison method failed: %s', ex, exc_info=ex)
 
-    def get_by_id(self, _id: str) -> ItemMeta:
+    def get_by_id(self, id_: str) -> ItemMeta:
         """Get an item by ID if it exists, or return None."""
         db = self._load_or_initialize()
-        return db.get(_id)
+        return db.get(id_)
 
     def find(self, to_match: Dict[str, Any]) -> ItemMetaGen:
         """
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -7,13 +7,13 @@ import math
 import os
 import os.path
 import re
-from typing import Any, Dict, Optional
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import simplejson as json
 
 from .. import utils
 from ..errors import DuplicateError, SearchError
-from .core import _Backend
+from .core import ItemMeta, ItemMetaGen, _Backend
 
 logger = logging.getLogger(__name__)
 
@@ -54,8 +54,13 @@ class JSONBackend(_Backend):
         Path to the happi config.
     """
 
-    def __init__(self, path: str, initialize: bool = False, cfg_path: Optional[str] = None):
-        self._load_cache = None
+    def __init__(
+        self,
+        path: str,
+        initialize: bool = False,
+        cfg_path: Optional[str] = None
+    ) -> None:
+        self._load_cache: Dict[str, ItemMeta] = None
         # Determine the cfg dir and build path to json db based on that unless we're initted w/o a config
         if cfg_path is not None:
             cfg_dir = os.path.dirname(cfg_path)
@@ -70,7 +75,7 @@ class JSONBackend(_Backend):
         """Clear the loaded cache."""
         self._load_cache = None
 
-    def _load_or_initialize(self):
+    def _load_or_initialize(self) -> Optional[Dict[str, ItemMeta]]:
         """Load an existing database or initialize a new one."""
         if self._load_cache is None:
             try:
@@ -83,7 +88,7 @@ class JSONBackend(_Backend):
         return self._load_cache
 
     @property
-    def all_items(self):
+    def all_items(self) -> List[ItemMeta]:
         """All of the items in the database."""
         json = self._load_or_initialize()
         return list(json.values())
@@ -111,7 +116,7 @@ class JSONBackend(_Backend):
         # Dump an empty dictionary
         self.store({})
 
-    def load(self):
+    def load(self) -> Dict[str, ItemMeta]:
         """Load the JSON database."""
         with open(self.path, 'r') as f:
             raw_json = f.read()
@@ -119,7 +124,7 @@ class JSONBackend(_Backend):
         # Allow for empty files to be considered valid databases:
         return json.loads(raw_json) if raw_json else {}
 
-    def store(self, db):
+    def store(self, db: Dict[str, ItemMeta]) -> None:
         """
         Stash the database in the JSON file.
 
@@ -147,13 +152,13 @@ class JSONBackend(_Backend):
                     # Release lock in filesystem
                     fcntl.flock(f, fcntl.LOCK_UN)
 
-    def _iterative_compare(self, comparison):
+    def _iterative_compare(self, comparison: Callable) -> ItemMetaGen:
         """
         Yields documents in which ``comparison(name, doc)`` returns `True`.
 
         Parameters
         ----------
-        comparison : callable
+        comparison : Callable
             A comparison function with a signature of (item_id, doc).
         """
 
@@ -168,12 +173,12 @@ class JSONBackend(_Backend):
             except Exception as ex:
                 logger.debug('Comparison method failed: %s', ex, exc_info=ex)
 
-    def get_by_id(self, id_):
+    def get_by_id(self, _id: str) -> ItemMeta:
         """Get an item by ID if it exists, or return None."""
         db = self._load_or_initialize()
-        return db.get(id_)
+        return db.get(_id)
 
-    def find(self, to_match):
+    def find(self, to_match: Dict[str, Any]) -> ItemMetaGen:
         """
         Find an instance or instances that matches the search criteria.
 
@@ -191,7 +196,14 @@ class JSONBackend(_Backend):
 
         yield from self._iterative_compare(comparison)
 
-    def find_range(self, key, *, start, stop=None, to_match):
+    def find_range(
+        self,
+        key: str,
+        *,
+        start: Union[int, float],
+        stop: Optional[Union[int, float]] = None,
+        to_match: Dict[str, Any]
+    ) -> ItemMetaGen:
         """
         Find an instance or instances that matches the search criteria, such
         that ``start <= entry[key] < stop``.
@@ -227,7 +239,12 @@ class JSONBackend(_Backend):
 
         yield from self._iterative_compare(comparison)
 
-    def find_regex(self, to_match, *, flags=re.IGNORECASE):
+    def find_regex(
+        self,
+        to_match: Dict[str, Any],
+        *,
+        flags=re.IGNORECASE
+    ) -> ItemMetaGen:
         """
         Find an instance or instances that matches the search criteria,
         using regular expressions.
@@ -249,7 +266,12 @@ class JSONBackend(_Backend):
 
         yield from self._iterative_compare(comparison)
 
-    def save(self, _id: str, post: Dict[str, Any], insert: bool = True):
+    def save(
+        self,
+        _id: str,
+        post: Dict[str, Any],
+        insert: bool = True
+    ) -> None:
         """
         Save information to the database.
 
@@ -291,7 +313,7 @@ class JSONBackend(_Backend):
                 except KeyError:
                     raise SearchError("No item found {}".format(_id))
 
-    def delete(self, _id: str):
+    def delete(self, _id: str) -> None:
         """
         Delete an item instance from the database.
 
--- a/happi/backends/multi_db.py
+++ b/happi/backends/multi_db.py
@@ -31,7 +31,7 @@ def prevent_duplicate_ids(fn):
 
 class MultiBackend(_Backend):
     """
-    Multi-file backend.
+    Multi-backend backend.
 
     Combines multiple backends, prioritizing database files in order
     of appearance.
@@ -93,7 +93,7 @@ class MultiBackend(_Backend):
 
         Yields
         ------
-        ItemMetaGen
+        Dict[str, Any]
             A generator of metadata documents
         """
         for bknd in self.backends:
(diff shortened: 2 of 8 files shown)


tangkong:
This was my take on the prompt, I'm interested to see if anyone has alternative ideas on how best to implements something like this.

happi/backends/json_db.py
```diff
@@ -168,12 +173,12 @@ def _iterative_compare(self, comparison):
             except Exception as ex:
                 logger.debug('Comparison method failed: %s', ex, exc_info=ex)
 
-    def get_by_id(self, id_):
+    def get_by_id(self, _id: str) -> ItemMeta:
```
klauer, line 176: `id_` was because the name clashed with Python's builtin `id()`, not so much naming it consistently with the database key

happi/backends/multi_db.py
```diff
@@ -0,0 +1,193 @@
...
+            if doc['_id'] not in unique_ids:
+                unique_ids.add(doc['_id'])
+                yield doc
+
+    return wrapper
+
+
+class MultiBackend(_Backend):
+    """
+    Multi-file backend.
```
klauer, line 34: Rather "multi-backend backend" as it's not strictly file-based

happi/backends/multi_db.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def save(self, _id, post, insert=True):
+        """The current implementation of this backend is read-only."""
+        raise NotImplementedError("The Questionnaire backend is read-only")
+
+    def delete(self, _id):
+        """The current implementation of this backend is read-only."""
+        raise NotImplementedError("The Questionnaire backend is read-only")
+
+    def get_by_id(self, id_: str) -> ItemMeta:
```
klauer, line 110:
```suggestion
    def get_by_id(self, id_: str) -> Optional[ItemMeta]:
```

happi/backends/multi_db.py
```diff
@@ -0,0 +1,193 @@
...
+        Returns the first match, in config priority
+
+        Parameters
+        ----------
+        id_ : str
+            id to search for
+
+        Returns
+        -------
+        ItemMeta
```
klauer, line 123:
```suggestion
        Dict[str, Any] or None
```

happi/client.py
```diff
@@ -801,10 +801,32 @@ def from_config(cls, cfg=None):
...
-        db_kwargs = cfg_parser['DEFAULT']
+        if len(cfg_parser.keys()) > 1:
+            # multi backend configuration have multiple sections
+            multi_backend = BACKENDS['multi']
+            sub_backends = []
+            for section in cfg_parser.keys():
+                bknd = cls._get_backend_from_config(cfg_parser[section], cfg)
+                sub_backends.append(bknd)
+            # move DEFAULT from index 0 to -1
+            default_backend = sub_backends.pop(0)
```
klauer, line 812: It's a bit arbitrary to have DEFAULT go to the end, as I think a similar argument could be made for it to be highest priority.

I think the as-defined order makes more sense and would vote to remove this line

klauer (commented):
Aside from the backend priorities and some nitpicks about docstrings, I think this looks really good. 👍

happi/backends/multi_db.py
```diff
@@ -0,0 +1,193 @@
...
+        ------
+        ItemMetaGen
+            A generator of metadata documents
+        """
+        for bknd in self.backends:
+            yield from bknd.find(*args, **kwargs)
+
+    def save(self, _id, post, insert=True):
+        """The current implementation of this backend is read-only."""
+        raise NotImplementedError("The Questionnaire backend is read-only")
```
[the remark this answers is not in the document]
klauer: Or perhaps the first writable database? Skipping the questionnaire, were it listed first, for example

happi/client.py
tangkong: My reasoning here was that smaller backends could essentially patch the DEFAULT backend.  In many cases the additional devices might not be unique, but slightly modified versions of devices existing in the DEFAULT backend.  In that case if DEFAULT was highest priority, the client wouldn't return the patched version.

Python's `configparser` also always pulls DEFAULT to be the first section listed, regardless of its placement in the cfg file itself, so that aspect of the ordering is lost.  Lacking a better way of reclaiming that order I thought it'd make sense to toss it at the end

happi/backends/json_db.py
tangkong: I was wondering about that.  Good to know

happi/backends/multi_db.py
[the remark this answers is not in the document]
ZLLentz: I think it's a bit more complicated than this.

If you save on an existing item, surely it should modify the database source that you loaded that item from?

If you create a new item, the first writable database seems reasonable, but I wonder if this ever gets us into trouble with people accidentally saving their updates in the wrong place. Is it possible that the API needs to be extended in some way to specify which database to include the post in?

Another way to approach this is to say "hey, use a different happi config when you want to make edits" but I wonder if this ends up being difficult.

When used in IPython for example this is pretty reasonable but not for a cli usage. Maybe the cli just needs special handling where it selects which specific backend to write to.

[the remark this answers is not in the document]
tangkong: I was unsure about this.  If we pick the first writabale / non-duplicate database, I'd be worried the user might not know where the Item went.  (since the backend already obfuscates which backend items are read from)

I'm ok implementing it that way, if we think that's acceptable.  I didn't think it'd be a big deal to make this backend read-only and have people make changes via a single-backend client.

happi/client.py
klauer: I guess I'm not sure why we'd have backends with the same (or very similar) information in them. I'd expect it to be used like "backend 1 has the hutch happi database" and "backend 2 has the photon side happi database" and "backend 3 has the accelerator side happi database", so priority wise search "hutch -> photon -> accelerator". 

Interesting about `DEFAULT` always coming first. `DEFAULT` coming first in configparser and last in happi feels weird, but perhaps it's just me. In that case, I'd say for the multi-backend scenario it should be recommended to _not_ use `DEFAULT` at all so you have clear control over the ordering.

happi/backends/multi_db.py
```diff
@@ -0,0 +1,193 @@
...
+
+from .core import ItemMeta, ItemMetaGen, _Backend
+
+logger = logging.getLogger(__name__)
+
+
+def prevent_duplicate_ids(fn):
+    """
+    Decorator that remembers which document _id's have been yielded and
+    prevents subsequent documents with matching _id's from being yielded
```
ZLLentz, line 17: Does this mean that the only way to "shadow"/replace an entry from another database is to cautiously make sure that the _id field matches exactly? Is there any way to do that in most of these backends? In the json backend you can hand-modify your file carefully I suppose.

It's worth noting that e.g. the "name" field doesn't have to match the "_id" field for happi to work.

happi/backends/multi_db.py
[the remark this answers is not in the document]
klauer: I can see the different scenarios for this and agree that it's not quite so clear-cut or obvious how this should be implemented.

happi/client.py
ZLLentz: > Python's configparser also always pulls DEFAULT to be the first section listed, regardless of its placement in the cfg file itself, so that aspect of the ordering is lost.

I didn't know this. I guess we shouldn't be using DEFAULT at all in these config files then, to preserve some sort of logical order?

I think by and large patching over the main db isn't the main use case of this, rather I'd expect the different dbs to be responsible for different subsets of the beamline (mostly)

happi/backends/multi_db.py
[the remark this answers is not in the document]
ZLLentz: I vote for no handling in this backend, and you can go call the specific backend you want to save to. The cli can have an extra arg for saving where you specify which config to save to.

happi/client.py
tangkong: > I guess I'm not sure why we'd have backends with the same (or very similar) information in them. 

The use case I had in mind was for something like XPP, where they can move devices between branches.  The information would be identical except for a few fields, so instead of changing the device_config they could just add a patching config. (if I understood the use case right)

> DEFAULT coming first in configparser and last in happi feels weird, but perhaps it's just me. 

Though the ordering is different, the behavior actually matches I think.  In both cases, fields in non-`DEFAULT` configs take precedence, and if there is a missing entry in one of the non-`DEFAULT` sections, the value from `DEFAULT` is used.  Phrase another way, `DEFAULT` never overrides information from a different section.  (the ordering of keys in the `ConfigParser` doesn't actually matter for the config files itself`)

ZLLentz: I guess this gives us a lot of flexibility over how to do mode swaps: shadowing, holding two configs that get swapped, etc.

happi/backends/multi_db.py
tangkong: Your understanding is right.  I wrote this thinking that `_id` was the unique field, is that not the case?  Or is it that the json-backend needs the `_id` to match the top level key?

happi/backends/multi_db.py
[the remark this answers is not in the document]
tangkong: No-writing for this backend seems good then.  We can tackle saving in the CLI separately, it's reasonable to list out the different backends for the user to pick from.

happi/backends/multi_db.py
ZLLentz: `_id` is the unique field but I don't quite remember if there's a way to control the value of it in every case

happi/client.py
tangkong: I think we can still use DEFAULT, but not using it would probably make the config priority clearer for users.

happi/backends/multi_db.py
ZLLentz: For example, what if I have a central mongodb + my personal little json file and I want to either shadow the existing items or I want to make sure I don't have an _id conflict- can I necessarily find a way to do these things for any backend?

happi/client.py
```diff
@@ -801,10 +801,32 @@ def from_config(cls, cfg=None):
         cfg_parser = configparser.ConfigParser()
         cfg_file = cfg_parser.read(cfg)
```
ZLLentz, line 802: A bit of a left field suggestion/idea: any thoughts on if we should be able to include paths to other HAPPI_CFG files within our happi config file? For example:

my_happi_config.cfg:

```
[MINE]
path=my_happi_cfg.json
[XPP]
cfg=xpp_config_file.cfg
```

xpp_config_file.cfg, in pink mode:

```
[XPP_HUTCH]
xpp_hutch_pink.json
[XPP_BEAMLINE]
xpp_beamline.json
```

and in mono mode, the hutch changes it to:

```
[XPP_HUTCH]
xpp_hutch_mono.json
[XPP_BEAMLINE]
xpp_beamline.json
```

happi/backends/multi_db.py
tangkong: I see, `_id` isn't really a field the user has direct control over while saving, etc.  

Maybe shadowing by name is a better choice here?

ZLLentz: I think I want to take a moment to read the code carefully and see who/when decides the id- this comment was triggered by checking our db and seeing that the ids and names don't quite match

klauer: I think we need to chat about `_id` one day soon. I'd rather not get into the details on this PR, though.

happi/client.py
klauer: Neat idea

happi/backends/multi_db.py
ZLLentz: I suspect maybe we set the id to the name and then never update it if the name changes which might be the more appropriate thing to address rather than the implementation here
https://github.com/pcdshub/device_config/blob/e9fcbc89e03c5b0e2ac20e6633c6f2ee726a1007/db.json#L3-L23

happi/client.py
tangkong: I'd personally want to keep everything in one HAPPI_CFG file, so I don't have to go searching for the other files being referenced.  But aside from that I don't really see an issue with this.

happi/backends/multi_db.py
ZLLentz: I think this PR does it correctly and I think there's some other bug in the code related to _id that let's our database look as it does for some entry

happi/client.py
ZLLentz: I guess the idea here was that XPP wouldn't need to change any global files to make other people's lightpath apps learn that the devices are all in the main line now

ZLLentz: Another way to accomplish the above is to just have XPP change xpp_devices.json as needed

happi/backends/multi_db.py
```diff
@@ -0,0 +1,193 @@
...
+    @prevent_duplicate_ids
+    def find(self, *args, multiples=False, **kwargs) -> ItemMetaGen:
+        """
+        Find an instance or instances that matches the search criteria.
+
+        Parameters
+        ----------
+        multiples : bool
+            Find a single result or all results matching the provided
+            information.
```
ZLLentz, line 90: I think this keyword arg is actually unused here

tangkong: It's unused in all the other backends too.  I thought about implementing it here, but forgot to.  Is it something we care about?  I think we can just remove it for consistency.

happi/client.py
tangkong: A quick update to the config parsing logic to make the DEFAULT section optional

happi/backends/multi_db.py
ZLLentz: > It's unused in all the other backends too

Somehow I didn't notice, whoops! Maybe a cleanup problem for another time. It seems like the other backends just give multiples and don't define the kwarg at all...

ZLLentz (approved):

klauer (approved):

tangkong:
Thanks for the reviews, some good discussion here.  I'll make a few followup issues shortly

PR #289, opened 2022-10-20
DOC: v2.0.0

ZLLentz:
## Description

Run the release notes script and clean up the output a bit (consistency, categorization)
Reviews requested from contributors

## Motivation and Context

I want to tag all of our packages
v2.0.0 = API break (remove deprecated methods)

## How Has This Been Tested?

CI only

This is documentation

## Pre-merge checklist
Checklist isn't relevant for this one


ZLLentz:
Somehow this got posted while I was writing the description, which has been lost. Whoops!

docs/source/releases.rst
```diff
@@ -2,6 +2,59 @@ Release History
...
+- For happi load, fall back to Python REPL if IPython is not available.
+- Added MultiBackend, which allows a happi Client to serve information
+  from multiple databases simultaneously.  Updates config parsing logic
+  to match.
+- Added ``happi audit`` function for running checks on happi database items.
+- Restored --json option for ``happi search``.
+
+Bugfixes
+--------
+- Removed extraneous extraneous print in `happi load`.
```
klauer, line 41: Is single quote italic and double code in rst? I get this confused too easily...

```suggestion
- Removed extraneous extraneous print in ``happi load``.
```

klauer (approved):
v2.0.0 - hard to believe it's here already.

docs/source/releases.rst
ZLLentz: Refreshing my memory- single quote italic hits the sphinx "default_role" thing, unless that is unset, in which case it does a block.
This repo has a default role set, so you're right.
https://github.com/pcdshub/happi/blob/d822eb50920c126737089bccd7448b91506bebf7/docs/source/conf.py#L99

ZLLentz:
> We should update the documentation to include details on the new features

You're right, we didn't make any commits in the docs at all since the last tag. I can double-back and give a try at it myself in this PR after I push all the other repos closer to tag. I don't think we'll push the env until next week in any case.

JJL772 (approved):

tangkong:
> You're right, we didn't make any commits in the docs at all since the last tag. I can double-back and give a try at it myself in this PR after I push all the other repos closer to tag. I don't think we'll push the env until next week in any case.

I think most of the docs stuff might fall on me, we'll see who gets to it first I guess

ZLLentz:
Let's merge this and tag now then, we can upload improved docs after

PR #292, opened 2022-11-15
MAINT: scripted fix for precommit

ZLLentz:
PR generated from script to fix .pre-commit-config.yaml in all repos, switching out the missing flake8 gitlab mirror for github.


ZLLentz:
I think this is correct but there is additional to do- perhaps in another PR? Something has broken.

tangkong:
Failure is due to something in line-profiler changing, rooted in pcdsutils' Line profiler functions.  see https://github.com/pcdshub/pcdsutils/pull/68

tangkong (approved):

PR #294, opened 2022-11-18
ENH: add delete cli command

### LINKED ISSUE #293: CLI: additional operations, delete, copy?
## Expected Behavior
We could probably rather easily add some cli commands to copy entries to a new name or delete items

## Current Behavior
These commands don't exist

## Possible Solution
An additional cli command

N/A

## Context
Helping some users with happi and ran into this 

## Your Environment
pcds-5.5.1

tangkong:
## Description
Adds a delete cli command.  Touches up some of the `add --clone` documentation to hopefully make it more obvious you can copy items. 

Modified existing test db to have `_id` match `name`, because happi was getting confused.  (happi client assumes `_id_key = name`)

## Motivation and Context
I was asked about this twice in one week, meaning it was time to make it easier for people

closes #293 

## How Has This Been Tested?
Added a delete test.  Looked at the `happi -h` output

This PR, docstrings 

## Pre-merge checklist

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -291,6 +291,18 @@ def add(ctx, clone: str):
 @happi_cli.command()
 @click.argument('name', type=str)
 @click.pass_context
+def copy(ctx, name):
+    """
+    Copy the item NAME.
+
+    Simply wraps ``happi add --clone``
+    """
+    ctx.invoke(add, clone=name)
+
+
+@happi_cli.command()
+@click.argument('name', type=str)
+@click.pass_context
 def delete(ctx, name: str):
     """
     Delete an existing entry.  Only accepts exact names
@@ -302,7 +314,9 @@ def delete(ctx, name: str):
         raise click.ClickException(f'Could not find item ({name}): {e}')
 
     item.show_info()
-    if click.confirm('Are you sure you want to delete this entry?'):
+    if click.confirm('Are you sure you want to delete this entry? \n'
+                     'Remember you can mark an entry as inactive with \n'
+                     '"happi edit my_device_name active=false"'):
         logger.info('Deleting item')
         client.remove_item(item)
         click.echo(f'Entry {name} removed')
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -237,7 +237,7 @@ def search_parser(
               'Provide the name of the item to clone.')
 @click.pass_context
 def add(ctx, clone: str):
-    """Add new entries interactively."""
+    """Add new entries interactively. Copy entries with --clone."""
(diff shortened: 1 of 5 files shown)


tangkong:
Noooooo the profiler

happi/cli.py
```diff
@@ -237,7 +237,7 @@ def search_parser(
               'Provide the name of the item to clone.')
 @click.pass_context
 def add(ctx, clone: str):
-    """Add new entries interactively."""
+    """Add new entries or copy existing entries."""
```
klauer, line 240: Is it worth having an additional top-level alias `copy` (= `add --clone`) to avoid confusion?

happi/cli.py
```diff
@@ -288,6 +288,28 @@ def add(ctx, clone: str):
...
+    Delete an existing entry.  Only accepts exact names
+    """
+    client = ctx.obj
+    try:
+        item = client.find_item(name=name)
+    except SearchError as e:
+        raise click.ClickException(f'Could not find item ({name}): {e}')
+
+    item.show_info()
+    if click.confirm('Are you sure you want to delete this entry?'):
```
ZLLentz, line 305: Small suggestion: it might be appropriate to let the user know that they can mark items as inactive instead of deleting them. I know sometimes you absolutely want to nuke an entry but a lot of the time I think people will stumble upon the delete command and not know you can do:

```
happi edit my_device_name active=false
```

happi/cli.py
tangkong: I was looking into a way to make click redirect a `happi copy` command to the `happi add --clone` help message, but got side tracked.  I like the idea

ZLLentz (approved):

happi/cli.py
tangkong: added in [660ed1c](https://github.com/pcdshub/happi/pull/294/commits/660ed1cc4e85227617b635628a9a0edc061d66dd)

happi/cli.py
tangkong: added `happi copy` in [660ed1c](https://github.com/pcdshub/happi/pull/294/commits/660ed1cc4e85227617b635628a9a0edc061d66dd)

ZLLentz:
I like how succinct the copy command can be here.

Note that the behavior is probably slightly different than what a user might expect copy to, since clone drops you into an interactive prompt session with new defaults.

tangkong:
I thought about making a command where it just copies everything but the name, but felt like making users acknowledge exactly what's going on is ok.  Not the most script-friendly, but that's what the python client is for right?...

tangkong:
Merging this so I can avoid merge conflicts on another PR

PR #295, opened 2022-11-19
ENH: add audit functions - arg/kwarg information verification and wait for connection

### LINKED ISSUE #290: ENH: additional audit functions, wait_for_connection, check args/kwargs defined
## Expected Behavior
`happi audit` added some basic audit functions, which are a good starting point but could use some expansion. 

Checking that args/kwargs are defined if referenced
- If an arg / kwarg is requested but unfilled in the container, happi will substitute an empty string.  Devices with this issue can be instantiated, but might not throw an error at creation. 

wait_for_connection audit
- An extension of the above, perhaps we can take more time and perform a full connection test on all the components

## Current Behavior
These audit functions don't exist

## Possible Solution
Additional functions!

## Context
Ken ran into a misconfigured at1k4 / at1k4_calc

## Your Environment
pcds-5.5.0

tangkong:
## Description
Adds additional audit functions for:
* ensuring any jinja-filled args or kwargs have corresponding entries
* checking that all signals in a device are connected

## Motivation and Context
These got prototyped during some debugging and I figured it was a good time to push this forward

The args and kwargs test needs to pull the document from the client directly, since when a HappiItem is created, any fields that exist on the container but not in the document are filled with `None`

closes #290 

## How Has This Been Tested?
A test case for the args/kwargs audit has been added.  (this took the bulk of the time, the tests were finicky)

This PR

## Pre-merge checklist

### DIFF
--- a/happi/audit.py
+++ b/happi/audit.py
@@ -6,8 +6,9 @@ an Exception with a helpful error message.  These exception messages will be
 caught and organized by the cli audit tool.
 """
 import inspect
-import re
 from typing import Callable, List, Optional, Tuple
+
+from jinja2 import DebugUndefined, Environment, meta
 
 from happi import SearchResult
 
@@ -44,8 +45,7 @@ def check_extra_info(
 
     extra = result.item.extraneous.copy()
     for key in ignore_keys:
-        if key in extra:
-            extra.pop(key)
+        extra.pop(key, None)
     if len(extra) != 0:
         raise ValueError(f'Un-enforced metadata found: {list(extra.keys())}')
 
@@ -72,15 +72,14 @@ def check_wait_connection(result: SearchResult) -> None:
     assert (hasattr(dev, 'wait_for_connection') and
             callable(getattr(dev, 'wait_for_connection'))), \
            ('device has no wait_for_connection method, and is likely '
-            'not an ophyd device')
+            'not an ophyd v1 device')
 
     try:
         dev.wait_for_connection(timeout=5)
     except TimeoutError as te:
-        # If we encounter a timeout error, gather some more details stats
-        if hasattr(dev, "walk_signals"):
-            sigs = list(sig.item for sig in dev.walk_signals())
-            conn_sigs = sum(sig.connected for sig in sigs)
+        # If we encounter a timeout error, gather some more detailed stats
+        sigs = list(sig.item for sig in dev.walk_signals())
+        conn_sigs = sum(sig.connected for sig in sigs)
 
         raise ValueError(f'{conn_sigs}/{len(sigs)} signals connected. \n'
                          f'original error: {te}')
@@ -99,20 +98,15 @@ def check_args_kwargs_match(result: SearchResult) -> None:
     cl = result.client
     doc = cl.find_document(**{cl._id_key: result[cl._id_key]})
     # pick out any jinja-like template
-    jinja_pattern = re.compile(r'^\{\{([a-zA-Z][a-zA-Z0-9_]*)\}\}$')
+    env = Environment(undefined=DebugUndefined)
 
-    for arg in doc.get('args', []):
-        patt_result = jinja_pattern.findall(arg)
-        if patt_result:
-            assert patt_result[0] in doc, \
-                   (f'{patt_result[0]} not found in item to fill args: '
-                    f'{doc["args"]}')
+    # render template and check if any variables were undefined
+    template = env.from_string(str(doc))
+    rendered = template.render(**doc)
+    undefined = meta.find_undeclared_variables(env.parse(rendered))
 
-    for kw_key, kw_value in doc.get('kwargs', {}).items():
-        if jinja_pattern.findall(kw_value):
-            assert kw_key in doc, \
-                   (f'{kw_key} not found in item to fill kwargs: '
-                    f'{doc["kwargs"]}')
+    assert len(undefined) == 0, \
+           f'undefined variables found in document: {undefined}'
 
 
 def verify_result(
--- a/happi/audit.py
+++ b/happi/audit.py
@@ -69,6 +69,11 @@ def check_wait_connection(result: SearchResult) -> None:
     """
     dev = result.get()
 
+    assert (hasattr(dev, 'wait_for_connection') and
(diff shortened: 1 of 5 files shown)


happi/audit.py
```diff
@@ -59,6 +61,55 @@ def check_name_match_id(result: SearchResult) -> None:
...
+def check_wait_connection(result: SearchResult) -> None:
+    """
+    Check if all the signals on the device can connect to their PV's
+
+    This can take a long time, consider filtering your query before running
+    """
+    dev = result.get()
+
+    try:
+        dev.wait_for_connection(timeout=5)
```
klauer, line 73: `wait_for_connection` is an ophyd-(presumably only v1)-ism. `happi audit` could potentially be used by others that don't use happi for ophyd devices. 

Should we have a `hasattr` check for `wait_for_connection`, or is there a better way to deselect this check for those users?

tangkong: Good point.  That was my reasoning for leaving out my lightpath audit function earlier, but I failed to remember that here.  I do think the `wait_for_connection` audit will be important for us to have deployed so I do want to include it.

My initial impression is that checking the database for ophyd-compatible devices would be a time-consuming process, since that's only known after instantiation.  I also wouldn't want to do something like abort the audit process if the method doesn't exist, since happi could host a mixture of device types. 

I think it's reasonable to simply let this check fail if the method doesn't exist (though it does that already, just not gracefully)  [d6074a4](https://github.com/pcdshub/happi/pull/295/commits/d6074a43559a53c048af99cb25e82c0f644a94b3)

tangkong:
Whoops I forgot this was still open.

happi/audit.py
```diff
@@ -59,6 +61,60 @@ def check_name_match_id(result: SearchResult) -> None:
...
+    except TimeoutError as te:
+        # If we encounter a timeout error, gather some more details stats
+        if hasattr(dev, "walk_signals"):
+            sigs = list(sig.item for sig in dev.walk_signals())
+            conn_sigs = sum(sig.connected for sig in sigs)
+
+        raise ValueError(f'{conn_sigs}/{len(sigs)} signals connected. \n'
```
klauer, line 85: `conn_sigs` and `sigs` can be unbound here if `walk_signals` doesn't exist on `dev`

Should `walk_signals` existence be assumed, instead, as getting here would indicate an ophyd v1 device?

happi/audit.py
```diff
@@ -59,6 +61,60 @@ def check_name_match_id(result: SearchResult) -> None:
...
+
+    i.e.: makes sure there is information in the entry named "extra"
+          if the kwargs = {'extra': '{{extra}}'}
+    """
+    # Happi fills in values when search result is created, must pull
+    # raw document from client level
+    cl = result.client
+    doc = cl.find_document(**{cl._id_key: result[cl._id_key]})
+    # pick out any jinja-like template
+    jinja_pattern = re.compile(r'^\{\{([a-zA-Z][a-zA-Z0-9_]*)\}\}$')
```
klauer, line 102: This expression will work for all common cases, but it's important to remember that this is jinja syntax. There could be things such as:
`{{ a or "foobar" }}` or `{{ a | filter }}`

Jinja, fortunately, has built-in functionality for this: https://github.com/pcdshub/happi/blob/0cb79b5e1f6bb06e1c03e17e22fd316b755ac447/happi/loader.py#L56

klauer (commented):
Looks good overall, but there are a couple things to fix up prior to merging.

happi/audit.py
tangkong: Ah, good point.  I think it makes sense to assume ohpyd v1 at this point, if a timeout error is thrown then surely we've reached the `Device.wait_for_connection` method..

happi/audit.py
tangkong: Ah, I should have thought to look through jinja 🤦

ZLLentz (approved):
This looks good to me aside from the weird CI thing

tangkong:
My next happi task will be to do the gha migration, hopefully resolving some of the travis shenanigans

happi/audit.py
```diff
@@ -59,6 +61,54 @@ def check_name_match_id(result: SearchResult) -> None:
...
+
+    i.e.: makes sure there is information in the entry named "extra"
+          if the kwargs = {'extra': '{{extra}}'}
+    """
+    # Happi fills in values when search result is created, must pull
+    # raw document from client level
+    cl = result.client
+    doc = cl.find_document(**{cl._id_key: result[cl._id_key]})
+    # pick out any jinja-like template
+    env = Environment(undefined=DebugUndefined)
```
klauer, line 101: Neat - is this better or a more modern version compared to the previous implementation? https://github.com/pcdshub/happi/blob/0cb79b5e1f6bb06e1c03e17e22fd316b755ac447/happi/loader.py#L56 (Just curious about the differences, not attempting to hold up this PR)

happi/cli.py
```diff
@@ -982,6 +985,13 @@ def audit(
         print(check_pt)
         return
 
+    if details:
+        check_fns = [fn for fn in checks if details in fn.__name__]
+        for fn in check_fns:
+            click.echo(inspect.getsource(fn))
+
+        return
```
klauer, line 993: What makes one of these like `--details` return early but some others not? The idea being that we should only do `--audit --details` and then call `happi audit --check`?

klauer (approved):
Looks good to me 👍

happi/audit.py
tangkong: I struggle to remember (though it hasn't been that long...), but I think this undefined kwarg is needed to make the Environment keep track of all the undeclared variables.  The loader tackles each arg individually, and this audit function tries to grab all undeclared at once.

I'd expect the default `Undeclared` handler to work, but after some experimentation I found it didn't.  So here we are 🤷

happi/cli.py
tangkong: yep, the `--details` flag is meant to be similar to `--list` , to be used more as a diagnostic or in preparation before the audit is run.

tangkong:
Ignoring the CI strangeness for now, hoping GHA behaves better

PR #296, opened 2023-02-01
BLD: GHA migration

tangkong:
## Description
Runs pcds-python-migration tools to update the repository for github actions.

## Motivation and Context
Travis doesn't like us.  

I manually bumped isort in my local submodule, since there's a chance the issue gets resolved and we can pin to 5.12.0.  I can always submit a pr to pcds-cookiecutter though

## How Has This Been Tested?
Tests still pass locally, let's see if gha likes this

This PR

## Pre-merge checklist
- [x] Test suite passes on continuous integration (~~Travis CI~~ github actions)


klauer:
So the pykerberos build is failing on pip install, I think we could either:
1. Remove psdm_qs_cli from our dev requirements so the pip test suite skips those kerberos tests
2. Add an input "install-packages" to our shared CI jobs so we can install the required system libraries for the pip version to work. Conda doesn't have this problem because it distributes everything as a binary and doesn't require building

pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+
+[tool.setuptools_scm]
+write_to = "happi/_version.py"
+
+[tool.flake8]
+exclude = "docs/*,versioneer.py"
+
+[tool.setuptools.packages.find]
+where = [ ".",]
+include = [ "happi",]
```
klauer, line 33: Looks like this is still generated with a previous version of the migration tools; could you take a look at your checked out copy?
```suggestion
include = [ "happi*",]
```

tangkong: argh I swear I pulled before running this

tangkong: fixed in https://github.com/pcdshub/pcds-python-migration-tools/pull/6 👍

tangkong:
So installing wheel mutes the pip errors, but doesn't install the gssabi C libraries needed for pykerberos.  We could add a step to install non-python packages in our CI jobs, but for now skipping those tests is probably fine.  Provided we follow the checklist and test locally that test should get run... right?

conda-recipe/meta.yaml
```diff
@@ -1,47 +1,55 @@
...
-      - happi
-    commands:
-      - happi --help
+  commands:
+  - happi --help
+  imports:
+  - happi
+  requires:
+  - pytest
+  - ipython
```
klauer, line 45: We could add `line_profiler` here in the `test` section instead of in the workflow configuration
Similarly, we can add `line-profiler` to `dev-requirements.txt` for pip

I think it's probably a better pattern to follow in general...

conda-recipe/meta.yaml
```diff
@@ -1,47 +1,55 @@
...
-      - coloredlogs
-      - entrypoints
-      - jinja2
-      - prettytable
-      - simplejson
-    run-constrained:
-      - psdm_qs_cli >=0.3.1
-      - pymongo >=4.0.2
+  build:
+  - python >=3.6
```
klauer, line 21: Likely 3.9+ after this PR

klauer (commented):
I think this looks good. If tests are passing, I believe we'll be in good shape.

The line-profiler max version pin looks like it'll need a follow-up to address. It'd be nice if 3.11 passed alongside the rest, but it can wait.

We'll need to update the release notes (looks like we have pre-release notes in this repo like pcdsdevices) to indicate all these changes. I'll wait to approve until then.

conda-recipe/meta.yaml
tangkong: This one is weird since `line-profiler` is being pulled from `pcdsutils`, but isn't a firm requirement there either.  (we just try to import it and return if we can't)

I do think it's probably good to have an explicit requirement here since we're advertising the line-profiler functionality, but I could also see a world where we leave it in the job recipe.

klauer: Yeah, that's fair... 

Looking back a bit, I think the original reason was that the package name differs on conda-forge and pypi. For our Travis CI builds, we erroneously shared `dev-requirements.txt` between conda and pip (so if we included the dependency there, 1 of 2 always failed trying to install `line-profiler` or `line_profiler`). Now that we use the `test` section from the conda recipe, that limitation is no longer an issue.

.github/workflows/standard.yml
```diff
@@ -17,8 +17,8 @@ jobs:
       # Extras that will be installed for both conda/pip:
       testing-extras: "pcdsdevices pcdsutils"
       # Extras to be installed only for conda-based testing:
-      conda-testing-extras: "line_profiler<4.0.0"
+      conda-testing-extras: ""
       # Extras to be installed only for pip-based testing:
-      pip-testing-extras: "line-profiler<4.0.0 wheel"
+      pip-testing-extras: "wheel"
```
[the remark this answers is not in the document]
tangkong: Doesn't it? 
[new log](https://github.com/pcdshub/happi/actions/runs/4068015539/jobs/7006060077#step:12:135): creates wheel for sentinels
[old log](https://github.com/pcdshub/happi/actions/runs/4059930110/jobs/6988509834#step:12:147):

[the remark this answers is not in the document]
klauer: Check the step before it in 'new log', though: https://github.com/pcdshub/happi/actions/runs/4068015539/jobs/7006060077#step:11:340
(Perhaps that's a separate issue and means we need 'wheel' also in our pyproject.toml build section?)

conda-recipe/meta.yaml
tangkong: It's apparently still happening.  https://github.com/pcdshub/pcds-ci-helpers/blob/2844a430ae6db9c19d82a213ea8dfa76985d0730/.github/workflows/python-conda-test.yml#L45

tangkong: Fixed via https://github.com/pcdshub/pcds-ci-helpers/pull/99 🎉

.github/workflows/standard.yml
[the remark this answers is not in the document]
tangkong: If I understand this all correctly, wheel needs to be present before any pip installations occur in order to avoid this error.  Currently we install all the testing dependencies in one go, so we can: 
- install wheel in a prior step
  - in the ["Upgrade pip"](https://github.com/pcdshub/pcds-ci-helpers/blob/97cea13ebe9d3cd2db451571cd0eacd2f218b8f2/.github/workflows/python-pip-test.yml#L87) or ["build wheel and source distribution"](https://github.com/pcdshub/pcds-ci-helpers/blob/97cea13ebe9d3cd2db451571cd0eacd2f218b8f2/.github/workflows/python-pip-test.yml#L91) steps?  (these are the ones that made sense to me at first glance)
- install the test dependencies with the `--pep-517` pip option

All of this is driven by the impending [PEP517](https://peps.python.org/pep-0517/) and an effort to make all packages installable via wheels ([pip deprecation warning issue](https://github.com/pypa/pip/issues/8559))

I do think this can be pushed out of the scope of this PR.  I'll make an issue with the above text for the ci-helpers, since I'm not fully sure which approach to take

ZLLentz (approved):
I read through all the lines to make sure nothing weird made it in automatically, it all looks good to me

klauer (approved):
LGTM, let's get this in

Issue #297, opened 2023-02-10
Happi search didn't find all cxi devices


ZLLentz:
## Expected Behavior

`happi search name=cxi* lightpath=False` should show all cxi devices with lightpath=False

## Current Behavior

It doesn't find them all (???)

## Steps to Reproduce (for bugs)

1. happi search name=cxi* lightpath=False
2. note that kb1 is missing
3. happi search cxi_kb1_hx
4. note that lightpath=False here

## Context

Trying to make cxi devices load again

## Your Environment

pcds_conda 5.6.0

tangkong:
Apparently the default in the container does not properly carry through to the item in a way that `happi.Client.search_regex` recognizes (possibly other search methods as well)

There's no concise way to describe this but:
This script prints:
```python
from happi.client import Client
from happi.client import SearchResult
cl = Client(path='/cds/home/r/roberttk/test/cxi_db_230210.json')

all_cxi_false = set([r.metadata['name'] for r in cl.search_regex(name='cxi.*', lightpath='False')])
all_cxi = set([r.metadata['name'] for r in cl.search_regex(name='cxi.*')])

print('cxi_kb1_hx' in all_cxi)
print('cxi_kb1_hx' in all_cxi_false)
```

True, True for:
```
{
  "_id": "cxi_kb1_hx",
  "active": true,
  "args": [
    "{{prefix}}"
  ],
  "beamline": "CXI",
  "creation": "Tue Aug 21 09:15:22 2018",
  "device_class": "pcdsdevices.device_types.IMS",
  "kwargs": {
    "name": "{{name}}"
  },
  "last_edit": "Tue Aug 21 09:23:32 2018",
  "lightpath": false,       <<<<<<<<<<<<<<
  "macros": null,
  "name": "cxi_kb1_hx",
  "parent": null,
  "prefix": "CXI:KB1:MMS:05",
  "screen": null,
  "stand": "KB1",
  "system": null,
  "type": "pcdsdevices.happi.containers.LCLSItem",
  "z": -1
}
```

True, False for:
```
{
  "_id": "cxi_kb1_hx",
  "active": true,
  "args": [
    "{{prefix}}"
  ],
  "beamline": "CXI",
  "creation": "Tue Aug 21 09:15:22 2018",
  "device_class": "pcdsdevices.device_types.IMS",
  "kwargs": {
    "name": "{{name}}"
  },
  "last_edit": "Tue Aug 21 09:23:32 2018",
  "macros": null,
[...]

PR #300, opened 2023-02-21
ENH: `happi repair`

tangkong:
## Description
Adds the `happi repair` command, which (currently) aims to update the database to align with the information expected by the container

Some attempts are made to cancel the saving process safely, as well as prompt the user for missing information

## Motivation and Context

#297 , where initial search failures were tracked down 

A short list of cases I tried to cover / thought about:
|                      | Mandatory             | Optional                |
|----------------------|-----------------------|-------------------------|
| Missing, has default | not possible       | fill with default value |
| Missing, no default  | prompt                | fill with 'null'        |

Currently if you run the audit and there are no mandatory items with missing data, the procedure will just re-save the item.  I wanted to only save if there was a change made, but couldn't find a clean way to determine if this was the case without manipulating the backend data directly.  

## How Has This Been Tested?
Interactively, though I should probably write tests.

This PR, docstrings incoming

image

## Pre-merge checklist
- [x] Test suite passes on continuous integration (GHA)

### DIFF
--- a/happi/audit.py
+++ b/happi/audit.py
@@ -58,7 +58,8 @@ def check_name_match_id(result: SearchResult) -> None:
     """
     id_field = result.metadata.get('_id')
     name_field = result.metadata.get('name')
-    assert id_field == name_field, f'id: {id_field} != name: {name_field}'
+    if id_field != name_field:
+        raise ValueError(f'id: {id_field} != name: {name_field}')
 
 
 def check_wait_connection(result: SearchResult) -> None:
@@ -69,10 +70,10 @@ def check_wait_connection(result: SearchResult) -> None:
     """
     dev = result.get()
 
-    assert (hasattr(dev, 'wait_for_connection') and
-            callable(getattr(dev, 'wait_for_connection'))), \
-           ('device has no wait_for_connection method, and is likely '
-            'not an ophyd v1 device')
+    if not (hasattr(dev, 'wait_for_connection') and
+            callable(getattr(dev, 'wait_for_connection'))):
+        raise ValueError('device has no wait_for_connection method, '
+                         'and is likely not an ophyd v1 device')
 
     try:
         dev.wait_for_connection(timeout=5)
@@ -105,8 +106,8 @@ def check_args_kwargs_match(result: SearchResult) -> None:
     rendered = template.render(**doc)
     undefined = meta.find_undeclared_variables(env.parse(rendered))
 
-    assert len(undefined) == 0, \
-           f'undefined variables found in document: {undefined}'
+    if len(undefined) != 0:
+        raise ValueError(f'undefined variables found in document: {undefined}')
 
 
 def find_unfilled_mandatory_info(
@@ -141,8 +142,8 @@ def check_unfilled_mandatory_info(result: SearchResult) -> None:
     This may arise from manual editing of the backend database.
     """
     unfilled_info = find_unfilled_mandatory_info(result)
-    assert len(unfilled_info) == 0, \
-           f'unfilled mandatory information found: {unfilled_info}'
+    if len(unfilled_info) != 0:
+        raise ValueError(f'unfilled mandatory information found: {unfilled_info}')
 
 
 def verify_result(
--- a/happi/audit.py
+++ b/happi/audit.py
@@ -119,6 +119,19 @@ def find_unfilled_mandatory_info(
             if getattr(result.item, info) is None]
 
 
+def find_unfilled_optional_info(
+    result: SearchResult
+) -> list[str]:
+    """
+    Return all optional fields that are missing a value
+    """
+    # cannot getattr for extraneous info, shortcircuit conditional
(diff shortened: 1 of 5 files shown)


klauer:
Would you consider moving the "find items with missing (required) metadata" out to a reusable utility function (that this CLI tool would then use)? It might be nice to programmatically be able to do the same sort of things

happi/cli.py
```diff
@@ -1075,6 +1077,73 @@ def audit(
         click.echo(f'# devices failed: {len(unique_fails)} / {len(results)}')
 
 
+@happi_cli.command()
+@click.pass_context
+@click.option('--fix-optional/--ignore-optinal', 'fix_optional', default=False,
```
klauer, line 1082:
```suggestion
@click.option('--fix-optional/--ignore-optional', 'fix_optional', default=False,
```

happi/cli.py
```diff
@@ -1075,6 +1077,73 @@ def audit(
...
+        if fix_optional:
+            empty_opt = find_unfilled_optional_info(res)
+            req_info = itertools.chain(req_info, empty_opt)
+
+        try:
+            req_field = next(req_info)
+        except StopIteration:
+            req_field = False
+
+        res_name = res['_id']
```
klauer, line 1121: Seeing `_id` reminded me of something tangential to "missing info" but I think likely something `happi repair` could (should?) handle.

I think we've discussed and agreed (correct me if I'm misremembering) that `name` and `_id` should be identical. This tool could check if that's the case and fix it, replacing `_id` with the value in `name`.

Follow-up PR, perhaps?

happi/cli.py
```diff
@@ -1075,6 +1077,73 @@ def audit(
...
+
+        # fix each mandatory field
+        logger.info(f'repairing ({res_name})...')
+        while req_field:
```
klauer, line 1125: Why the fancy iterator usage?
Would `for req_field in req_info` not suffice?

happi/cli.py
```diff
@@ -1075,6 +1077,73 @@ def audit(
...
+                break
+
+        # re-save after creating container
+        try:
+            # will save optional data with defaults
+            # optional data without defaults saved as null
+            res.item.save()
+        except KeyboardInterrupt:
+            # Finish the current save if interrupted
+            logger.warning('caught keboard interrupt, finishing')
```
klauer, line 1142:
```suggestion
            logger.warning('caught keyboard interrupt, finishing')
```

klauer: Is the expectation that an interrupted `save()` could be repaired by trying again after `KeyboardInterrupt`?

klauer (approved):
Thanks for tackling this one!

Looks good to me overall - left a few thoughts in the comments as usual.

happi/cli.py
tangkong: Ah this was necessary before I remembered about the `prompt_for_entry` helper.  Simple will do here

happi/cli.py
tangkong: My experience showed that an interrupted `save()` may end up with an empty `db.json`, and saving again would write the file properly.  I'm not entirely sure that's the intention, but that worked for me.

happi/cli.py
tangkong: augh spelling is hard

happi/cli.py
klauer: Oh, interesting... That makes me think we should be a lot more careful with `save()` in general.

It might be good to suppress `KeyboardInterrupt` during `save()`, or have the backend do a safer "write-and-move" operation... Let's chat about this maybe tomorrow

happi/cli.py
tangkong: I think a follow-up PR would be a good place to address this.  I envision this tool growing as we find different things to fix.

happi/audit.py
```diff
@@ -109,6 +109,42 @@ def check_args_kwargs_match(result: SearchResult) -> None:
...
+def check_unfilled_mandatory_info(result: SearchResult) -> None:
+    """
+    Check that all mandatory EntryInfo have a value.
+
+    This identifies problems originating from a mismatch between
+    an item's backend representation and desired container.
+    This may arise from manual editing of the backend database.
+    """
+    unfilled_info = find_unfilled_mandatory_info(result)
+    assert len(unfilled_info) == 0, \
```
ZLLentz, line 144: It is possible to run Python in such a way that it ignores all assert statements, so if this is used in try/except blocks it may be better to raise some other exception directly

ZLLentz (approved):
I like the approach here :+1: this will be a good tool for saving us after container updates

happi/audit.py
tangkong: Oh TIL.  fixed in 38f77f3

tangkong:
As is tradition, the happi hydra spawns two issues with every one cut down.  I'm going to merge this and we can tackle followups in other PR's

PR #308, opened 2023-04-26
ENH: add post-md-set hook to loader

tangkong:
## Description
Adds a hook to allow code to be run on the object after the `.md` attribute is set.  This would allow the loaded object to interact with its metadata. 

## Motivation and Context
Most recently https://github.com/pcdshub/pcdsdevices/pull/1132/, but also the lightpath input/output branches setting in pcdsdevices previously [(link)](https://github.com/pcdshub/pcdsdevices/blob/02b295e236a21b4d8fcdc6969041f5b8a3b5fa03/pcdsdevices/interface.py#L1787)

Previously, we've hooked the `.md` setter to do some operations after happi attempts to set  the `.md` attribute.  This has caused (me) confusion, and unless thoroughly annotated the purpose of the setter is not clear (mostly to me with  my goldfish memory)

## How Has This Been Tested?
Added a quick test 

Here, and pre-release notes eventually

## Pre-merge checklist

### DIFF
--- a/happi/loader.py
+++ b/happi/loader.py
@@ -91,6 +91,7 @@ def fill_template(
 def from_container(
     item: HappiItem,
     attach_md: bool = True,
+    run_post_attach_hook: bool = True,
     use_cache: bool = True,
     threaded: bool = False,
 ) -> Any:
@@ -200,6 +201,12 @@ def from_container(
             setattr(obj, 'md', item)
         except Exception:
             logger.warning("Unable to attach metadata dictionary to device")
+
+        if run_post_attach_hook and hasattr(obj, 'post_happi_md'):
+            try:
+                obj.post_happi_md()
+            except Exception as ex:
+                logger.warning(f"Unable to run {obj}.post_happi_md: {ex}")
 
     # Store the device in the cache
     cache[item.name] = obj
--- a/happi/loader.py
+++ b/happi/loader.py
@@ -123,6 +123,9 @@ def from_container(
         The item to load.
     attach_md : bool, optional
         Attach the container to the instantiated object as ``md``.
+    run_post_attach_hook : bool, optional
+        Run ``post_happi_md`` method on the device after attaching the container
+        if possible.
     use_cache : bool, optional
         When devices are loaded they are stored in the ``happi.cache``
         dictionary. This means that repeated attempts to load the device will
--- a/happi/tests/test_loader.py
+++ b/happi/tests/test_loader.py
@@ -1,4 +1,5 @@
 from typing import Any
+from unittest.mock import patch
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


happi/loader.py
```diff
@@ -201,6 +205,12 @@ def create_arg(arg):
         except Exception:
             logger.warning("Unable to attach metadata dictionary to device")
 
+        if run_post_attach_hook and hasattr(obj, 'post_happi_md'):
+            try:
+                obj.post_happi_md()
```
ZLLentz, line 210: I think no args is the correct decision here
We could have made `md` the arg, but as it's not something we're introducing in this step I think it's correct to let the user find self.md if they want to access it in the hook.

One design question: should this still run if the md setattr fails? There's a check for this above.
I guess the state where someone has made a device that can't get an md setattr'd but decided to include a post-hook is pretty low

happi/loader.py
```diff
@@ -201,6 +205,12 @@ def create_arg(arg):
         except Exception:
             logger.warning("Unable to attach metadata dictionary to device")
 
+        if run_post_attach_hook and hasattr(obj, 'post_happi_md'):
```
ZLLentz, line 208: I like the method name except that it sounds super similar to the happi item's `post` method. Probably not an issue though. https://github.com/pcdshub/happi/blob/c46a5e9e81741b01e5f98629cae460662be33d1c/happi/item.py#L367

happi/tests/test_loader.py
```diff
@@ -149,3 +150,28 @@ def test_filter_kwargs(item_jinja: OphydItem):
     assert getattr(filtered_dev, 'blank_exclude', 'DNE') == 'DNE'
     for bl in blanks:
         assert getattr(filtered_dev, bl, 'DNE') == 'DNE'
+
+
+class PostDevice:
+    def post_happi_md(self):
+        print('post_happi_md hook run')
```
ZLLentz, line 157: Maybe this could have set an attribute e.g. `self.post_count += 1` and then there wouldn't have been a need to invoke patch/mock in the test cases.

ZLLentz (commented):
Some design considerations and missing pre-release here, but I think this is solid

happi/loader.py
tangkong: `after_happi_md` doesn't have the same ring to it..
`happi_md_hook`?  eh

happi/loader.py
ZLLentz: Makes sense to me, let's keep it as-is

happi/loader.py
ZLLentz: Yeah idk it shouldn't be confusing at all to people who haven't poked around with happi internals
I blame English for ascribing too many meanings to "post"

tangkong:
This dropped off my radar, whoops.  Added the docs

ZLLentz (approved):
I'm :+1: on this pr

PR #310, opened 2023-05-02
GUI/MNT: let HappiDeviceTreeView include items where key is not found

tangkong:
## Description
* Instead of omitting items, show them under a catch-all tree item "[KEY NOT FOUND]"

## Motivation and Context
If you're looking to group items by a metadata key, I think you'd expect every item in the database to be represented.  If you were to group the database by a field and then search for an item, it might just not exist if its metadata didn't feature that field.

In reality this was inspired by the logger spam when switching between views.  Currently ATEF is the only real user of this model view, I think this is a better way to handle missing metadata.  

Sidebar: I considered modifying `get_happi_entry_value` to pass null/None values from the database through, but in the end this can be grouped together with missing keys without loss of information.  I do think raising a ValueError there is a bit overkill, since happi containers missing a key is a bit overkill.  In the end this is the smallest change that solves my current problem.

## How Has This Been Tested?
interactively, primarily using the ATEF Happi viewer.

This PR

old: (item disappears when grouped)
image

new: 
image

## Pre-merge checklist
- [ ] New/changed functions and methods are covered in the test suite where possible

### DIFF
--- a/docs/source/upcoming_release_notes/310-mnt_missing_key.rst
+++ b/docs/source/upcoming_release_notes/310-mnt_missing_key.rst
@@ -1,0 +1,22 @@
+310 mnt_missing_key
+###################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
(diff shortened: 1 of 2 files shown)


ZLLentz (approved):
I like this approach, having an "other" section in the window view here makes a lot of sense.

PR #315, opened 2023-06-15
ENH: happi config cli sub-commands

untzag:
## Description

Adding new sub-command-group `config` to CLI.
- `happi config edit` - open config file in $EDITOR
- `happi config init` - create new config file with default options
- `happi config show` - show location & contents of config file

## Motivation and Context

See #314 

## How Has This Been Tested?

Tested on my machine.

CLI is auto-documented.

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/315-happi_config_cli_sub-commands.rst
+++ b/docs/source/upcoming_release_notes/315-happi_config_cli_sub-commands.rst
@@ -7,9 +7,9 @@ API Changes
 
 Features
 --------
-- happi config edit - open config file in $EDITOR
-- happi config init - create new config file with default options
-- happi config show - show location & contents of config file
+- CLI command ``happi config edit`` - open config file in $EDITOR
+- CLI command ``happi config init`` - create new config file with default options
+- CLI command ``happi config show`` - show location & contents of config file
 
 Bugfixes
 --------
--- a/docs/source/upcoming_release_notes/315-happi_config_cli_sub-commands.rst
+++ b/docs/source/upcoming_release_notes/315-happi_config_cli_sub-commands.rst
@@ -1,0 +1,24 @@
+315 happi config cli sub-commands
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- happi config edit - open config file in $EDITOR
+- happi config init - create new config file with default options
+- happi config show - show location & contents of config file
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- untzag
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -402,7 +402,7 @@ def load(ctx, item_names: list[str]):
 
     logger.debug('Starting load block')
     # retrieve client
-    client = ctx.obj
+    client = get_happi_client_from_config(ctx.obj)
 
     devices = {}
     names = " ".join(item_names)
@@ -1209,9 +1209,9 @@ def show():
 
     def draw_line():
         try:
-             click.echo("-"*os.get_terminal_size()[0])
+            click.echo("-"*os.get_terminal_size()[0])
         except OSError:
-             # non-interactive mode (piping results). No max width
+            # non-interactive mode (piping results). No max width
             click.echo("-"*79)
 
     draw_line()
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -1163,13 +1163,14 @@ def init(overwrite):
     # find config_filepath
     try:
         config_filepath = Path(happi.client.Client.find_config())
+    except OSError:
+        config_filepath = Path(platformdirs.user_config_dir("happi")) / "happi.cfg"
+    else:
         if not overwrite:
             click.echo("Found existing config file at:")
             click.echo(f"  {config_filepath}")
             click.echo("Stopping! Use --overwrite to destroy this config file.")
             return
-    except OSError:
-        config_filepath = Path(platformdirs.user_config_dir("happi")) / "happi.cfg"
     click.echo("Creating new config file at:")
     click.echo(f"  {config_filepath}")
 
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -1157,7 +1157,8 @@ def edit_config():
 @config.command()
 @click.option("--overwrite/--no-overwrite", "overwrite", default=False,
               help="Overwrite existing config.")
-def init(overwrite):
+@click.option("--backend", type=click.Choice(["json"], case_sensitive=False), default="json")
+def init(overwrite, backend):
     """Create configuration file with default options."""
 
     # find config_filepath
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -40,6 +40,19 @@ version_msg = f'Happi: Version {happi.__version__} from {happi.__file__}'
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
 
+def get_happi_client_from_config(path):
+    # gather client
+    try:
+        client = happi.client.Client.from_config(path)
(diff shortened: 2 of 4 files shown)


happi/cli.py
```diff
@@ -66,13 +67,15 @@ def happi_cli(ctx, path, verbose):
...
-    # User objects must be assigned to ctx.obj, which will be passed
-    # through to new context objects
-    ctx.obj = client
+    try:
+        client = happi.client.Client.from_config(cfg=path)
+        logger.debug("Happi client: %r" % client)
+        # insert items into context to be passed to subcommands
+        # User objects must be assigned to ctx.obj, which will be passed
+        # through to new context objects
+        ctx.obj = client
```
ZLLentz, line 76: Is there any way we can give a better error message for the user when there is no client but they invoke a routine that requires one? I assume this one throws a cryptic `AttributeError`, though I haven't run it myself yet.

ZLLentz (commented):
I haven't had time to run these yet but the code looks good
I have two minor thoughts that might end up as my/our todo items later

happi/cli.py
tangkong: My thought was initially to store the path in the `ctx.obj` attribute and call `Client.from_config` in each of the subcommands.  A bit more verbose but maybe less cryptic?

klauer: This is why I was saying each command that requires a _valid and working_ client should instantiate the client on its own.

That is to say, `happi search` (and all others that require the client) should have a `get_cli_happi_client_from_config(ctx)`.

Were it structured like this, then exceptions could be raised or caught in a sensible spot - rather than before control flow gets to the command the user is trying to run.

happi/cli.py
```diff
@@ -1134,7 +1137,12 @@ def repair(
             break
 
 
-@happi_cli.command()
+@click.group(help="Commands related to the happi config file.")
```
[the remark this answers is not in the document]
tangkong: Good point about the entrypoint names.  We could do something like this (with better names): 

```python
@click.group(...)
def _happi_cli():
    # rename the original happi_cli

def happi_cli():
    _happi_cli.add_command(config)
    _happi_cli()
```

I tried out `click.CommandCollection` briefly, but it breaks the context passing.  It's probably not too hard to get it to work, but the above solution is maybe simpler

docs/source/upcoming_release_notes/315-happi_config_cli_sub-commands.rst
```diff
@@ -0,0 +1,24 @@
...
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- happi config edit - open config file in $EDITOR
+- happi config init - create new config file with default options
+- happi config show - show location & contents of config file
```
klauer, line 12:
```suggestion
- CLI command ``happi config edit`` - open config file in $EDITOR
- CLI command ``happi config init`` - create new config file with default options
- CLI command ``happi config show`` - show location & contents of config file
```

happi/cli.py
```diff
@@ -1146,6 +1154,60 @@ def edit_config():
...
+@config.command()
+@click.option("--overwrite/--no-overwrite", "overwrite", default=False,
+              help="Overwrite existing config.")
+def init(overwrite):
+    """Create configuration file with default options."""
+
+    # find config_filepath
+    try:
+        config_filepath = Path(happi.client.Client.find_config())
+        if not overwrite:
```
klauer, line 1166: Would you mind putting this `if` statement in an `else:` block to reduce the scope of the `try/except OSError`?

happi/cli.py
```diff
@@ -1146,6 +1154,60 @@ def edit_config():
...
+            click.echo(f"  {config_filepath}")
+            click.echo("Stopping! Use --overwrite to destroy this config file.")
+            return
+    except OSError:
+        config_filepath = Path(platformdirs.user_config_dir("happi")) / "happi.cfg"
+    click.echo("Creating new config file at:")
+    click.echo(f"  {config_filepath}")
+
+    # find database_filepath
+    database_filepath = Path(platformdirs.user_data_dir("happi")) / "db.json"
```
klauer, line 1177: This assumes the JSON backend, which is generally fine. I think for future-proofing, it might be wise to add a `backend` argument here that defaults to `"json"` and raises `NotImplementedError` if it's not `"json"`.

happi/cli.py
tangkong, line 1140: `happi -h` only shows `config` as a separate command, hiding the subcommands.  `happi config -h` shows those subcommands.  I think this is probably ok, as we're not exactly advertising these to the average user.  

If we wanted to be more transparent, we could list the subcommand names in the help message, but I don't think that's strictly necessary

tangkong (approved):
This looks good to me on the whole.  Works for me interactively.  I think Zach's points are worth addressing, but I have no complaints otherwise

happi/cli.py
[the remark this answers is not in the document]
untzag: @ZLLentz nice for compatibility with what? `happi edit-config` was only merged into master last week, and has never been released. I don't think it's a priority to retain that spelling as an option.

I'm probably missing something here.

happi/cli.py
untzag: Great idea. Something I should be more in the habit of doing.

happi/cli.py
untzag: Seems like a wise idea to have the option `--json`. By implementing it as a `click.Choice` users will never be able to enter anything except the options given (currently only `json`).

My "hot take" is that it's probably never going to be easy to just guess what config should look like for database backends... too many site-specific choices are inevitable. But perhaps it could go down a different codepath where it asked some questions.

happi/cli.py
untzag: I agree that it's good enough, but will change if there's somebody with a strong desire.

happi/cli.py
```diff
@@ -1146,6 +1154,60 @@ def edit_config():
...
+
+    click.echo("Done!")
+
+
+@config.command()
+def show():
+    """Show configuration file in current state."""
+    config_filepath = Path(happi.client.Client.find_config())
+    click.echo(f"File: {config_filepath}")
+    click.echo("-"*79)
```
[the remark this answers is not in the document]
untzag: That's pretty :rose:, done.

happi/cli.py
untzag: Do you want me to do this in this PR, or should we merge this and then do that refactor in a separate PR? I'm okay either way... really up to y'all.

tangkong: Do you mind doing that in this PR?  I think it should be a quick enough addition

tangkong:
Whoops, the failures from the test suite are coming from the way I assumed the context would be loaded.  Sorry, I didn't realize this refactor would be so involved.  The change needed is pretty quick though:  Also use the `happi_cfg` fixture and pass that in as the `obj` in any cli test.  

An Example:
``` python
def test_search(client: happi.client.Client, happi_cfg: str):
    res = client.search_regex(beamline="TST")

    with search.make_context('search', ['beamline=TST'], obj=happi_cfg) as ctx:
        res_cli = search.invoke(ctx)

    assert [r.item for r in res] == [r.item for r in res_cli]
```

I can push some changes real quick

klauer (approved):
I think this looks good now. 

Thanks for the useful contribution and working with us on the suggestions! 👍

tangkong (approved):
Looks good to me as well, appreciate you fielding all our requests 👍

happi/cli.py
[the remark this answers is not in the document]
ZLLentz: e.g. I need to modify an environment or create a new one to pick up the new entrypoint definition if I want to test the code here

ZLLentz (approved):

PR #317, opened 2023-06-29
DOC/REL: v2.3.0

ZLLentz:
## Description

- Run the release notes script
- Hand-edit a bit
- Remove some characters from older docs that were showing up as invalid on windows

Before merging I want to double-check the github actions docs build to make sure I didn't mess up the rst syntax

## Motivation and Context

- There are some good additions recently from @untzag that should be included in a tag


tangkong (commented):
I downloaded the docs from the last job and it looks good to me.  (Though the `platformdirs` hyperlink doesn't seem to work, I believe it needs an [underscore](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages) )
image

Every time I work on docs I have to look up all the rst syntax again too 😞

ZLLentz:
my brain did not see the underscore when I checked the rst docs yesterday
one more edit coming right up

tangkong (approved):
As the googlers say, LGTM

PR #327, opened 2023-08-22
ENH: Match Name to ID Field 

### LINKED ISSUE #302: ENH: add name = _id fix to `happi repair`
## Expected Behavior
`happi repair` should also check and align an item's `name` with their `_id`.   

The rationale behind this should also be documented once we agree on it, since we keep questioning why we hold this convention.  a quick trawl through issues / pr's shows the reason to be...?
- `_id` is the unique key in the json database, and should match `name` to make searching in either direction

Some quick bits of conversation: 
- https://github.com/pcdshub/happi/pull/287#discussion_r989473921
- https://github.com/pcdshub/happi/issues/286#issuecomment-1249842505
- https://github.com/pcdshub/happi/pull/260

There are probably other checks to add, maybe they get lumped in with this?

## Current Behavior
`happi repair` only checks that all mandatory information have values.  

## Possible Solution
Additional code block in the `happi repair` command
## Context
brought up in #300 

## Your Environment
pcds-5.6.0

laura-king:
## Description

When running happy repair, will check if name and id fields are matching. If not, will set the name field to the id field. 

Added comparison and attribute setting in repair function. Added click context-based unit test for setting name and id equivalence. 

Changed 'res_name' variable to 'res_id' to represent correct value.

## Motivation and Context

Closes #302 

## How Has This Been Tested?

Unit test using click contexts that call repair on the bad-happi-config.

Here

## Pre-merge checklist
- [ ] Test suite passes on GitHub Actions

### DIFF
--- a/docs/source/upcoming_release_notes/302-repair_match_name_to_id.rst
+++ b/docs/source/upcoming_release_notes/302-repair_match_name_to_id.rst
@@ -1,0 +1,22 @@
+302 repair_match_name_to_id
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Issue 302: Add functionality to happi 'repair' that ensures that the name and id fields of a device are the same. 
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- laura-king
--- a/docs/source/upcoming_release_notes/302-repair_match_name_to_id.rst
+++ b/docs/source/upcoming_release_notes/302-repair_match_name_to_id.rst
@@ -11,7 +11,7 @@ Features
 
 Bugfixes
 --------
-- Issue 302: Add functionality to happi 'repair' that ensures that the name and id fields of a device are the same. 
+- Issue 302: Add functionality to happi 'repair' that ensures that the name and id fields of a device are the same.
 
 Maintenance
 -----------
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -1149,7 +1149,7 @@ def repair(
         # check name and id parity
         if res['name'] != res_id:
             # set name to match id
-            setattr(res.item, 'name', res_id)
+            res.item.name = res_id
 
         # re-save after creating container
         try:
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -1153,15 +1153,20 @@ def repair(
         if fix_optional:
             req_info.extend(find_unfilled_optional_info(res))
 
-        res_name = res['_id']
+        res_id = res['_id']
 
         # fix each mandatory field
-        logger.info(f'repairing ({res_name})...')
+        logger.info(f'repairing ({res_id})...')
         for req_field in req_info:
             info = res.item._info_attrs[req_field]
             req_value = prompt_for_entry(info)
             info.enforce_value(req_value)
             setattr(res.item, req_field, req_value)
+
+        # check name and id parity
+        if res['name'] != res_id:
+            # set name to match id
+            setattr(res.item, 'name', res_id)
 
         # re-save after creating container
         try:
--- a/happi/tests/test_cli.py
+++ b/happi/tests/test_cli.py
@@ -97,6 +97,23 @@ def test_cli_no_argument(runner: CliRunner):
     assert 'Commands:' in result.output
 
 
+def test_repair_name(runner: CliRunner, bad_happi_cfg: str):
+    search_string = '_id=tst_id'
+
+    # make sure the name and id are different
+    with search.make_context('search', [search_string], obj=bad_happi_cfg) as ctx:
+        res = search.invoke(ctx)
+    assert all([r['_id'] != r['name'] for r in res])
+
+    # run repair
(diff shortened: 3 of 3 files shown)


happi/cli.py
```diff
@@ -1136,16 +1136,21 @@ def repair(
...
         for req_field in req_info:
             info = res.item._info_attrs[req_field]
             req_value = prompt_for_entry(info)
             info.enforce_value(req_value)
             setattr(res.item, req_field, req_value)
 
+        # check name and id parity
+        if res['name'] != res_id:
+            # set name to match id
+            setattr(res.item, 'name', res_id)
```
klauer, line 1152:
```suggestion
            res.item.name = res_id
```
`setattr` is good for the dynamic case - where you don't know the attribute name - but since we do here we may as well use the normal assignment operator

docs/source/upcoming_release_notes/302-repair_match_name_to_id.rst
```diff
@@ -0,0 +1,22 @@
...
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Issue 302: Add functionality to happi 'repair' that ensures that the name and id fields of a device are the same.
```
klauer, line 14: See our `pre-commit` output here in the checks? https://github.com/pcdshub/happi/actions/runs/5942583794/job/16115855304
It's verifying some basic style stuff and it found an issue on this line I'm commenting on - turns out you have a bit of whitespace at the end of the line.

`pre-commit` will actually fix this automatically for you, if you run it locally.
Now, you can certainly run pre-commit stuff on-demand, but I find it helpful to fully install it (by way of `pre-commit install`) so it runs each time before you've made a commit.

Our hooks are defined at the repo top level in `.pre-commit-config.yaml`
Here's a link to the usage: https://pre-commit.com/#usage and if you scroll up to the top, there's some useful background information there, too.

klauer (commented):
I think this looks good and implements exactly what the issue states. Nicely done!

Now only I'm wondering if the original issue had `name` and `_id` reversed? That is to say, we probably want to sync "_id" to "name" instead of vice-versa (as in this PR). `name` is very easy to update, but `_id` is not. We don't have a tool for renaming or moving entries to a different `_id` to my recollection.

Anyway, I'll let Zach/Robert weigh in here before we go ahead...

tangkong:
The main caveat there (I think) is that a working db will have no conflicting id's, but could have conflicting names. So setting id <- name might cause a collision.

But Ken's point is still true, the opposite (id - > name) gives us less control

happi/tests/test_cli.py
```diff
@@ -100,6 +100,23 @@ def test_cli_no_argument(runner: CliRunner):
     assert 'Commands:' in result.output
 
 
+def test_repair_name(runner: CliRunner, bad_happi_cfg: str):
+    search_string = '_id=tst_id'
+
+    # make sure the name and id are different
+    with search.make_context('search', [search_string], obj=bad_happi_cfg) as ctx:
+        res = search.invoke(ctx)
+    assert all([r['_id'] != r['name'] for r in res])
```
tangkong, line 109: This confused me since there's only one item in the `bad_happi_cfg` where the id and name don't match, but the test is totally still valid so I can't complain too much

tangkong (commented):
I think this LGTM, pending the pre-commit fixes.  I realized that in the case where there's two items with the same `_id`, one of them will just get nuked by this.  That's not the fault of this PR, but something that might want to be addressed.  

Or possibly we choose to table this since we're hoping to swap over to the mongoDB anyway 🤷 (where we can set a unique key and guarantee a field like `_id` remains unique)

tangkong:
Ah I also just noticed you pushed to the pcdshub repo instead of your own fork.  We typically abide by the fork-branch-pullrequest model, as it keeps our repos clean and works better with some of our CI jobs.

It's ok this time, but next time you should make changes to a branch on your own fork, then submit a pull request to merge that fork (e.g. `laura-king:repair-name`) into `pcdshub:master`

docs/source/upcoming_release_notes/302-repair_match_name_to_id.rst
ZLLentz: I think this is the only active problem with this PR

ZLLentz (approved):

ZLLentz:
> I realized that in the case where there's two items with the same _id, one of them will just get nuked by this.

I don't think this is possible in the json backend, unless you hand-edit the files

klauer:
> I don't think this is possible in the json backend, unless you hand-edit the files

Right, because the JSON looks like this:

```
"_id_name": {
  "_id": "_id_name",
  "name": "ophyd_name"
},
```
(Making up an `_id` of `"_id_name"`, the key of the item dictionary has to match the `"_id"` entry in the key. `"name"` can be different, but we force it to be the same as in this PR.

happi/tests/test_cli.py
klauer: The search as Laura mentioned in our dev zoom makes it so it picks the invalid item - so `all()` here is really only checking that one

PR #330, opened 2023-09-01
Remove temporary file in JSON backend store function

### LINKED ISSUE #304: Clean up temporary file from #303
## Expected Behavior

The temporary file should be cleaned up if it can be.

## Current Behavior

The file is never cleaned up.
#303 was written assuming that the file could be useful, but I'm so sure any more.

## Possible Solution

Try/except clean up the file.
This doesn't need to be particularly robust, it just need to be a best-effort cleanup. Leaving the file around is not fatal.
As a bonus, the filename of the temp file can be simplified to just a hash.

laura-king:
## Motivation and Context

Closes #304  

If an error causes the temporary file in the json backend to not get copied properly, it would previously remain in the directory. It has not been needed so it will now be removed if left behind. 

In addition, because this file is no longer needed, the datetime and username are no longer needed in the file name. It is now just the unique hash to prevent collisions. Imports previously used to add time and username are now removed from the file.

## How Has This Been Tested?

Interactively, and a new test in test-backends.py 

Issues #303, #304

Comments within function itself and test. 

## Pre-merge checklist

### DIFF
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -150,6 +150,7 @@ class JSONBackend(_Backend):
             # remove temporary file
             if os.path.exists(temp_path):
                 os.remove(temp_path)
+            raise
 
     def _temp_path(self) -> str:
         """
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -2,14 +2,12 @@
 Backend implemenation using the ``simplejson`` package.
 """
 import contextlib
-import getpass
 import logging
 import math
 import os
 import os.path
 import re
 import shutil
-import time
 import uuid
 from typing import Any, Callable, Optional, Union
 
@@ -151,7 +149,6 @@ class JSONBackend(_Backend):
             logger.debug('JSON db move failed: %s', ex, exc_info=ex)
             # remove temporary file
             if os.path.exists(temp_path):
-                print(f"we are removing {temp_path}")
                 os.remove(temp_path)
 
     def _temp_path(self) -> str:
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -144,8 +144,6 @@ class JSONBackend(_Backend):
                 shutil.copymode(self.path, temp_path)
             shutil.move(temp_path, self.path)
         except BaseException as ex:
-
-            # log that the db move was not executed properly
             logger.debug('JSON db move failed: %s', ex, exc_info=ex)
             # remove temporary file
             if os.path.exists(temp_path):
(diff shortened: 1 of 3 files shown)


happi/backends/json_db.py
```diff
@@ -138,27 +136,30 @@ def store(self, db: dict[str, ItemMeta]) -> None:
...
+        except BaseException as ex:
 
-        if os.path.exists(self.path):
-            shutil.copymode(self.path, temp_path)
-        shutil.move(temp_path, self.path)
+            # log that the db move was not executed properly
+            logger.debug('JSON db move failed: %s', ex, exc_info=ex)
+            # remove temporary file
+            if os.path.exists(temp_path):
+                os.remove(temp_path)
```
klauer, line 152: After this line, we should re-raise the exception so that the caller can handle it, that is:
```python
                os.remove(temp_path)
            raise
```

klauer (commented):
Minor tweak, otherwise LGTM 👍

tangkong (approved):
seems good to me, pending Ken's suggestion

happi/backends/json_db.py
ZLLentz: I think this is especially important here because we're catching `BaseException` which might be e.g. a `KeyboardInterrupt`

happi/tests/test_backends.py
```diff
@@ -203,6 +203,25 @@ def test_json_tempfile_uniqueness():
...
 
+def test_json_tempfile_remove(monkeypatch):
+    # Set consistent temppath
+    jb = JSONBackend("testing.json", initialize=False)
+    temp_path = jb._temp_path()
+    jb._temp_path = lambda: temp_path
+
+    # Ensure file is created, then throw error
+    def shutil_move_patch(*args, **kwargs):
+        assert os.path.isfile(os.path.join(os.getcwd(), temp_path))
```
klauer, line 214: We should be a bit careful here - this assertion might raise, and it'd be caught by `with pytest.raises(BaseException)` below. Assertions are `AssertionError(Exception)` which inherits from `BaseException` (standard library has a nice ASCII tree of this [here](https://docs.python.org/3/library/exceptions.html#exception-hierarchy))

The short of it is: I think you should switch the below clause on line 220 to be `with pytest.raises(RuntimeError)` to make sure we're catching line 215 and not line 214

klauer (commented):
Looking good! One minor thing and then I think it's ready to go

happi/tests/test_backends.py
laura-king: This makes sense, and I'll update the raise. 

I will look more into how assert works in pytest-for some reason I thought the test would just count as a fail and move on. Not sure why I thought of it as isolated

ZLLentz (approved):
I think this is good to merge

klauer (approved):
LGTM 👍

PR #332, opened 2023-09-05
Import changes for pcdsutils and pcdsdevices

### LINKED ISSUE #331: Move github workflow dependencies to dev-requirements and conda recipe test dependencies
## Expected Behavior

In general, github workflow dependencies should be used sparingly (for things ONLY needed in the CI) because otherwise they obfuscate true package requirements.

Test requirements belong in two places:
https://github.com/pcdshub/happi/blob/861496e5b40e7ad2fc4be6bdbfb0a53073b4247b/conda-recipe/meta.yaml#L42-L45
https://github.com/pcdshub/happi/blob/861496e5b40e7ad2fc4be6bdbfb0a53073b4247b/dev-requirements.txt

## Current Behavior

pcdsdevices and pcdsutils are included in the workflow only: https://github.com/pcdshub/happi/blob/861496e5b40e7ad2fc4be6bdbfb0a53073b4247b/.github/workflows/standard.yml#L19

## Possible Solution

Move them over

laura-king:
## Description

Add pcdsdevices and pcdsutils to dev requirements and conda recipe, and remove it from the github actions extra testing-specific requirements. 

## Motivation and Context

The libraries should be considered required instead of optional for testing.
Closes #331 
inspired by #329 

## How Has This Been Tested?

Test suite runs, github actions workflow and fresh environment building completes as expected.

Here, and the aforementioned issues.

## Pre-merge checklist

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -27,8 +27,6 @@ requirements:
   - coloredlogs
   - entrypoints
   - jinja2
-  - pcdsutils
-  - pcdsdevices
   - platformdirs
   - prettytable
   - simplejson
@@ -45,6 +43,8 @@ test:
   - pytest
   - ipython
   - line_profiler
+  - pcdsutils
+  - pcdsdevices
 
 about:
   dev_url: https://github.com/pcdshub/happi/
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -27,6 +27,8 @@ requirements:
   - coloredlogs
   - entrypoints
   - jinja2
+  - pcdsutils
+  - pcdsdevices
   - platformdirs
   - prettytable
   - simplejson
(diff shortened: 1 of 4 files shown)


conda-recipe/meta.yaml
```diff
@@ -27,6 +27,8 @@ requirements:
   - coloredlogs
   - entrypoints
   - jinja2
+  - pcdsutils
```
klauer, line 30: It's not obvious where to put these without this extra info:
* The `requirements` section here is largely equivalent to `requirements.txt`(i.e., what you need to run the package in general)
* The `test` / `requires` section below is roughly equivalent to `dev-requirements.txt` (i.e., what you need to run the package test suite *in addition to* the run requirements above)
* `docs-requirements.txt` doesn't have an equivalent in the conda recipe, because we build our documentation in a pip-based job

That is to say, these 2 packages should go down to the `dev-requirements` equivalent section (test/requires below)

laura-king: Oh okay. Makes sense, will fix.

klauer (approved):
Looks right to me - thanks for the fix! 👍

tangkong (approved):

ZLLentz (approved):
LGTM!

PR #335, opened 2023-09-18
BUG: fix happi transfer default handling

### LINKED ISSUE #334: BUG: `happi transfer` default values are broken
## Expected Behavior
This should work...

## Current Behavior
* as expected list entries (input_branches, output_branches) are broken
* Taking default actually writes the whole string `take default: {value}`, instead of just `value`  ([src](https://github.com/pcdshub/happi/blob/064d4e79809214bd466922e0cf9cf68f59d7e308/happi/prompt.py#L185))
* when asked to amend entries, no option to take default. Also enforce regex rules can be hard to decipher

## Possible Solution
pls fix

Try to transfer from a a container with fewer fields to one with more

## Context
Tried to transfer from OphydItem -> LCLSItem and was thoroughly embarrassed 

## Your Environment
pcds-5.7.3

tangkong:
## Description
Fix happi transfer default handling (don't past default help text)
adds a test to try and cover this 

Also fixes `conftest.trim_split_output` because it just.. entirely didn't work.  More shame

In a burst of extra effort, disables assertion checks against expected stdout-style output.  The check still takes place, but simply prints (visible with `pytest -s` flag) if a mismatch is found.  In place of this, the happi_cfg is checked to verify the requested action has actually taken place.

Tests modified are:
- `test_transfer_cli`
- `test_transfer_cli_more`
- `test_search_json`  (this one is comparing to json, so I still assert the stdout comparisons here)
- `test_add_cli`
- `test_add_clone`

## Motivation and Context
closes #334 , in part (the non-list / dict comprehension parts)

## How Has This Been Tested?
Adds a test

This PR

## Pre-merge checklist

### DIFF
--- a/happi/prompt.py
+++ b/happi/prompt.py
@@ -200,7 +200,7 @@ def transfer_container(client, item, target):
         except TransferError as e:
             print(e)
             fix_prompt = f'New value for "{e.key}"'
-            d = getattr(getattr(target, e.key), 'default')
+            d = getattr(target, e.key).default
             new_val = click.prompt(fix_prompt, default=d)
             edits.update({e.key: new_val})
 
--- a/happi/prompt.py
+++ b/happi/prompt.py
@@ -183,8 +183,10 @@ def transfer_container(client, item, target):
                           f'entry "{nt}"')
         d = getattr(getattr(target, nt), 'default')
         val = click.prompt(missing_prompt, default=f'take default: {d}')
-        if val != 'take default':
+        if not ('take default' in val):
             edits.update({nt: val})
+        else:
+            edits.update({nt: d})
 
     # Actually apply changes and cast item into new container
     # Enforce conditions are dealt with here
@@ -198,7 +200,8 @@ def transfer_container(client, item, target):
         except TransferError as e:
             print(e)
             fix_prompt = f'New value for "{e.key}"'
-            new_val = click.prompt(fix_prompt)
+            d = getattr(getattr(target, e.key), 'default')
+            new_val = click.prompt(fix_prompt, default=d)
             edits.update({e.key: new_val})
 
     if not new_kwargs:
--- a/happi/prompt.py
+++ b/happi/prompt.py
@@ -6,6 +6,7 @@ import click
 import prettytable
 
 from happi.errors import EnforceError, TransferError
+from happi.item import EntryInfo
 from happi.utils import (OptionalDefault, is_valid_identifier_not_keyword,
                          optional_enforce)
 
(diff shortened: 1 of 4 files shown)


happi/prompt.py
```diff
@@ -198,7 +200,8 @@ def transfer_container(client, item, target):
         except TransferError as e:
             print(e)
             fix_prompt = f'New value for "{e.key}"'
-            new_val = click.prompt(fix_prompt)
+            d = getattr(getattr(target, e.key), 'default')
```
klauer, line 203:
```suggestion
            d = getattr(target, e.key).default
```

klauer (approved):
I think this looks good - a couple style suggestions is all

happi/prompt.py
```diff
@@ -183,8 +183,10 @@ def transfer_container(client, item, target):
                           f'entry "{nt}"')
         d = getattr(getattr(target, nt), 'default')
         val = click.prompt(missing_prompt, default=f'take default: {d}')
-        if val != 'take default':
+        if not ('take default' in val):
```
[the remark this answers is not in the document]
ZLLentz: Another way to handle this is to include the "default value is x" sort of information in the prompt and make the default value just the value

happi/prompt.py
ZLLentz: I was going to suggest the same thing

happi/tests/test_cli.py
```diff
@@ -44,14 +44,14 @@ def trim_split_output(strings: str, delim: str = '\n'):
     """
     date_pattern = r"\[(\d{4})[-](0[1-9]|1[012])[-].*\]"
 
-    ok_substrs = [r"happi.item."]
+    omit_substrs = [r"happi.containers.", r"- DEBUG -"]
```
ZLLentz, line 47: I'm not sure I actually understand what this function is trying to accomplish, so it's also hard to understand the edits

happi/tests/test_cli.py
```diff
@@ -60,19 +60,26 @@ def trim_split_output(strings: str, delim: str = '\n'):
...
 
     trimmed_output = trim_split_output(result.output)
     for message, expected in zip(trimmed_output, expected_output):
-        assert message == expected
+        if raise_failures:
+            assert message == expected
+        else:
+            if not message == expected:
+                print('Output does not match expected value:\n'
+                      f'- Output: ({message}\n- Expected: {expected}')
```
ZLLentz, line 82: I think the split here is a good call, it could be helpful debug information when the actual check fails

ZLLentz (approved):

happi/tests/test_cli.py
tangkong: There was apparently some output that had datetime information, as well as debug prints.  This function removes those from the observed output so the tests aren't time-dependent.

The containers bit is to remove containers collected from entrypoints.  (eg pcdsdevices).  this is admittedly not very robust, since external containers aren't guaranteed to be in a submodule containing 'happi.containers.'  
(in happi add we list all the available containers, and if you run this test suite in psbuild it'll find all the pcdsdevices containers)

happi/prompt.py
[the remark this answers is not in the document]
tangkong: Poking at this I remember that this was done this way to allow `default=None`.  There's certainly a better way about this either way though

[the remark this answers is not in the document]
tangkong: There indeed was, I wrote it a while back in `prompt_for_entry`.  🤦

PR #353, opened 2025-01-23
BLD: split ui dependencies into subpackage

### LINKED ISSUE #340: GUI / BLD: Make separate happi-qt sub-package?
## Expected Behavior
We might want to consider having a `happi[core]` and `happi[qt]` , once we start to accrue significant qt elements.

## Current Behavior
Qt widgets and their dependencies are grouped with the core code.  

## Possible Solution
[optional dependencies install instructions](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies)

## Context
#246 

## Your Environment
pcds-5.8.0

### LINKED ISSUE #347: BLD: consider backend specific package dependencies
## Expected Behavior
If a user wants to use the mongo backend for happi, we should supply a happi[mongo] subpackage etc

## Current Behavior
We don't install any backend specific dependencies, the only one that comes bundled is the json built-in

## Possible Solution
pip extras
conda outputs (once we actually figure out how this works, with lightpath as a guinea pig)

N/A

## Context
A thought that came up.  An extension of #340 

## Your Environment
pcds-5.8.4

tangkong:
## Description
Splits happi gui dependencies into subpackages (pip) or separate outputs (conda)

## Motivation and Context
closes #347 
closes #340 

## How Has This Been Tested?
Added some ui tests, but mostly relying on CI here to assess failure

This PR

## Pre-merge checklist
- [ ] Ran ``docs/pre-release-notes.sh`` and created a pre-release documentation page
- [ ] Pre-release docs include context, functional descriptions, and contributors as appropriate

### DIFF
--- a/gui-requirements.txt
+++ b/gui-requirements.txt
@@ -1,0 +1,2 @@
+PyQt5
+qtpy
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,11 +43,17 @@ file = [ "requirements.txt",]
 [tool.setuptools.dynamic.optional-dependencies.gui]
 file = "gui-requirements.txt"
 
+[tool.setuptools.dynamic.optional-dependencies.mongo]
+file = "mongo-requirements.txt"
+
 [tool.setuptools.dynamic.optional-dependencies.test]
-file = ["dev-requirements.txt", "gui-requirements.txt"]
+file = ["dev-requirements.txt", "gui-requirements.txt", "mongo-requirements.txt"]
 
 [tool.setuptools.dynamic.optional-dependencies.doc]
 file = "docs-requirements.txt"
+
+[tool.setuptools.dynamic.optional-dependencies.all]
+file = ["gui-requirements.txt", "mongo-requirements.txt"]
 
 [tool.pytest.ini_options]
 addopts = "--cov=."
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,8 +40,11 @@ content-type = "text/markdown"
 [tool.setuptools.dynamic.dependencies]
 file = [ "requirements.txt",]
 
+[tool.setuptools.dynamic.optional-dependencies.gui]
+file = "gui-requirements.txt"
+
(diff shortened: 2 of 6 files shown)


tangkong:
In the lightpath, we only adjusted the pip builds in the repo, and left the conda outputs for the feedstock.  I'm going to verify that the pip builds work here, then try applying the same conda output recipe format to the conda recipe here.  I think it'd be nice to have a build check independent of conda-forge, but I'm not sure if any of the output setup is conda-forge specific

tangkong:
A note: CI should remain unchanged as long as the `.[test]` dependency set remains unchanged.

gui-requirements.txt
```diff
@@ -0,0 +1,2 @@
+PyQt5
```
ZLLentz, line 1: There's at least some potential consideration for not having PyQt5 here, I don't remember best practices for qtpy apps

pyproject.toml
```diff
@@ -40,11 +40,20 @@ content-type = "text/markdown"
 [tool.setuptools.dynamic.dependencies]
 file = [ "requirements.txt",]
 
+[tool.setuptools.dynamic.optional-dependencies.gui]
+file = "gui-requirements.txt"
+
+[tool.setuptools.dynamic.optional-dependencies.mongo]
+file = "mongo-requirements.txt"
```
ZLLentz, line 47: I think this split is pretty reasonable :+1:

ZLLentz (commented):
Looks reasonable
Missing release notes

gui-requirements.txt
tangkong: .sWe include `PyQt5` for both `atef` and `lightpath`, which I was looking at for reference.  You're right that `qtpy` is supposed to be our qt-agnostic translation layer, though without it pip won't install PyQt5.  (at least from what I can verify from my testing)

ZLLentz (approved):

janeliu-slac (approved):

PR #357, opened 2025-02-24
MNT: make happi.cli.search_parser always return a list

### LINKED ISSUE #354: happi.cli.search_parser has misleading annotations
It claims to always return a `list`, but it does not. Sometimes it returns a set. We should fix the annotations and/or make this consistent.

tangkong:
## Description
- Makes the return type for `happi.cli.search_parser` always a list.  Previously this could sometimes return a `Set`
- Runs through a variety of type hinting problems revealed by mypy
- Fixes a straggling reference to `happi.containers.registry._registry`, uses the public API instead (`happi.containers.registry[]`)
- Makes a set -> list conversion in `happi.cli.load` more concise

## Motivation and Context
Closes #354 

I ran `mypy` along the way and ended up doing a bunch of other random things.  Including
- Not overwriting variables when changing types

## How Has This Been Tested?
Makes sure the test still ran, plus some interactive testing

This PR

## Pre-merge checklist

### DIFF
--- a/happi/cli.py
+++ b/happi/cli.py
@@ -156,6 +156,11 @@ def search_parser(
         regex matching.
     search_criteria : list of str
         The user's search selection from the command line.
+
+    Raises
+    ------
+    click.ClickException
+        Thrown if search criteria are invalid.
     """
     # Get search criteria into dictionary for use by client
     client_args = {}
@@ -232,7 +237,7 @@ def search_parser(
         final_results = range_list
     elif range_list and regex_list:
         # find the intersection between regex_list and range_list
-        final_results = set(range_list) & set(regex_list)
+        final_results = list(set(range_list) & set(regex_list))
     else:
         logger.debug('No regex or range items found')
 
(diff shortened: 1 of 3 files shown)


happi/cli.py
```diff
@@ -191,7 +196,7 @@ def search_parser(
                     range_list = new_range_list
                 else:
                     # subsequent ranges, only take intersection
-                    range_list = set(new_range_list) & set(range_list)
+                    range_list = list(set(new_range_list) & set(range_list))
```
ZLLentz, line 199: super nit: maybe we can hold off on listifying the range_list (and rename it to range_set?) until we return it. It probably doesn't actually matter but it seems wasteful to convert this to a list here and back to a set for the final_results in line 240.

ZLLentz (dismissed):
I'm on board with these changes, they solve the linked issues fully.

happi/cli.py
tangkong: I actually think we have to list-ify it here early, since there's a condition where we return `range_list` without modification.  (in the case where there is no `regex_list` to consider)

ZLLentz: I mean, we could `return list(range_set)` instead

tangkong: Ah yes, fair enough.  I'll be less lazy here

ZLLentz (approved):

PR #358, opened 2025-04-15
MNT: Adding CODEOWNERS file

tangkong:
(autogenerated) adding codeowners file


tangkong:
updating pre-commit while I'm here, because why not

Issue #299, opened 2023-02-17
Invalid containers can break `dict(client)` call


klauer:
## Current Behavior
Calling `dict(client)` when an invalid container is in the database raises an exception (and, of course, stops it from returning even a partial dictionary).

```python
$ python -m whatrecord.plugins.happi
Traceback (most recent call last):
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.6.0/lib/python3.9/site-packages/happi/client.py", line 305, in _get_item_from_document
    item = self.create_item(doc["type"], **doc)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.6.0/lib/python3.9/site-packages/happi/client.py", line 239, in create_item
    raise TypeError(
TypeError: The container class 'pcdsdevices.happi.containers.Leviton' is not in the registry

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.6.0/lib/python3.9/site-packages/happi/client.py", line 491, in __getitem__
    item = self._get_item_from_document(self.backend.get_by_id(key))
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.6.0/lib/python3.9/site-packages/happi/client.py", line 307, in _get_item_from_document
    raise EntryError(
happi.errors.EntryError: The information relating to the container class has been modified to the point where the object can not be initialized, please load the corresponding document

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.6.0/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.6.0/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.6.0/lib/python3.9/site-packages/whatrecord/plugins/happi.py", line 400, in <module>
    _cli_main()
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.6.0/lib/python3.9/site-packages/whatrecord/plugins/happi.py", line 393, in _cli_main
    results = main(**vars(args))
[...]

PR #362, opened 2025-06-03
BUG/MNT: return InvalidResult if container is malformed

tangkong:
## Description
Creates an InvalidResult that may be returned if a container is malformed.  Exceptions are still raised if keys do not exist.  This does change the behavior of a few methods, which used to simply skip results that couldn't be created.

This could be considered an API break.

Fixes a variety of type hinting errors along the way

## Motivation and Context
#299 

Also I wanted to test pcds-6.0.1 development

## How Has This Been Tested?
Mostly through test suite

This PR

## Pre-merge checklist

### DIFF
--- a/happi/client.py
+++ b/happi/client.py
@@ -10,6 +10,7 @@ import re
 import sys
 import time as ttime
 from collections.abc import Sequence
+from pathlib import Path
 from typing import Any, Optional, Union
 
 import platformdirs
@@ -68,7 +69,6 @@ class SearchResult(collections.abc.Mapping):
         return self._item
 
     def get(self, attach_md=True, use_cache=True, threaded=False):
-        '''(get) ''' + from_container.__doc__
         if self._instantiated is None:
             self._instantiated = from_container(
                 self.item, attach_md=attach_md, use_cache=use_cache,
@@ -810,7 +810,7 @@ class Client(collections.abc.Mapping):
         return _id
 
     @classmethod
-    def from_config(cls, cfg=None):
+    def from_config(cls, cfg: Optional[Union[str, os.PathLike]] = None):
         """
         Create a client from a configuration file specification.
 
@@ -836,6 +836,10 @@ class Client(collections.abc.Mapping):
         if not cfg:
             cfg = cls.find_config()
 
+        if cfg is None:
+            raise RuntimeError("No config found or provided")
+
+        cfg = Path(cfg)
         if not os.path.exists(cfg):
             raise RuntimeError(f'happi configuration file not found: {cfg!r}')
 
@@ -872,8 +876,8 @@ class Client(collections.abc.Mapping):
     @staticmethod
     def _get_backend_from_config(
         cfg_section: configparser.SectionProxy,
-        cfg: str
-    ) -> _Backend:
+        cfg: Union[str, os.PathLike]
+    ) -> Optional[_Backend]:
         # If a backend is specified use it, otherwise default
         if 'backend' in cfg_section:
             db_str = cfg_section.pop('backend')
--- a/happi/client.py
+++ b/happi/client.py
@@ -10,7 +10,7 @@ import re
 import sys
 import time as ttime
 from collections.abc import Sequence
-from typing import Any, Optional
+from typing import Any, Optional, Union
 
 import platformdirs
 
@@ -100,6 +100,34 @@ class SearchResult(collections.abc.Mapping):
 
     def __hash__(self) -> int:
         return hash((self.client.backend, self['_id']))
+
+
+class InvalidResult(collections.abc.Mapping):
+    """
+    A result from Client.search that could not be resolved to a valid HappiItem.
+    """
+
+    def __init__(self, doc: dict[str, Any], exc: Exception):
+        self.metadata = doc
+        self.exception = exc
+
+    def __getitem__(self, item):
+        return self.metadata[item]
+
+    def __iter__(self):
+        yield from self.metadata
+
+    def __len__(self):
+        return len(self.metadata)
+
+    def __repr__(self):
+        return (
+            f'{self.__class__.__name__}(metadata={self.metadata}, '
+            f'exception={type(self.exception)})'
+        )
+
+
+GenericResult = Union[SearchResult, InvalidResult]
 
 
 class Client(collections.abc.Mapping):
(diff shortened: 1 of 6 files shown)


docs/source/upcoming_release_notes/362-bug_dict_invalid.rst
```diff
@@ -0,0 +1,23 @@
+362 bug_dict_invalid
+####################
+
+API Breaks
+----------
+- Client search methods will now return an `InvalidResult` if a container
+  (e.g. `Happifitem`) cannot be created.
```
ZLLentz, line 7: happi fight em
```suggestion
  (e.g. `HappiItem`) cannot be created.
```

docs/source/upcoming_release_notes/362-bug_dict_invalid.rst
```diff
@@ -0,0 +1,23 @@
+362 bug_dict_invalid
+####################
+
+API Breaks
+----------
+- Client search methods will now return an `InvalidResult` if a container
```
ZLLentz, line 6: Docs suggestion: should we also document the previous behavior and rationale here for an API break?

> Previously, client search methods would raise a `KeyError` from the inciting exception. This `KeyError` will no longer be raised, and the inciting exception will be included as a data member in the `InvalidResult` object. This means downstream code that iterates over many `happi` entries doesn't need special error handling on big searches unless specifically the specific search item it needs is invalid.

happi/client.py
```diff
@@ -484,14 +512,21 @@ def validate(self):
...
-        return [res.item for res in self.search()]
+        return [res.item for res in self.search()
+                if isinstance(res, SearchResult)]
 
     def __getitem__(self, key):
         """Get an item ID."""
+        item_doc = self.backend.get_by_id(key)
+
+        if item_doc is None:
+            raise KeyError(key)
```
ZLLentz, line 523: Comment: It's good to preserve this error :+1:

happi/client.py
```diff
@@ -102,6 +102,34 @@ def __hash__(self) -> int:
         return hash((self.client.backend, self['_id']))
 
 
+class InvalidResult(collections.abc.Mapping):
```
ZLLentz, line 105: Question/Documentation: I tried to understand the code to document the practical similarities and differences between an `InvalidResult` and a `SearchResult`:

Similarities:
- You can still check metadata with key-based access (if it is present)
- Basic operations such as iteration, length checks still work (and will skip any of the missing data members)

Differences:
- There is no expectation that an `InvalidResult` will have well-formed data, some of it may be well-formed but data types are not guaranteed and some required data may be missing.
- You can't meaningfully compare two `InvalidResult` objects with equality checks.
- `InvalidResult` objects can't find the `item` that generated them via an `item` property (`SearchResult supports this)
- `InvalidResult` objects don't have a `get` function and cannot be used to instantiate real devices

Is this all correct? I think these are sensible limitations. Should any of these be added to `InvalidResult`?

happi/client.py
```diff
@@ -627,7 +664,7 @@ def search_regex(
         items = self.backend.find_regex(kwargs, flags=flags)
         return self._get_search_results(items)
 
-    def export(self, path=sys.stdout, sep='\t', attrs=None):
+    def export(self, path=sys.stdout, sep='\t', attrs: Optional[list[str]] = None):
```
ZLLentz, line 667: Question pre-dating PR: why does this function exist? Does it have utility over the json backend?

happi/client.py
```diff
@@ -645,6 +682,10 @@ def export(self, path=sys.stdout, sep='\t', attrs=None):
         devs = self.all_items
         logger.info('Creating file at %s ...', path)
         # Load item information
+        if attrs is None:
+            # TODO: decide how to handle None case, currently no-op
+            attrs = []
```
ZLLentz, line 687: Comment pre-dating PR: possibly this should an include-all-attrs instead if we care to support this function.
The fix here is better than the status quo where the function clearly errors out if you don't provide the `attrs` argument.

happi/client.py
```diff
@@ -831,8 +876,8 @@ def from_config(cls, cfg=None):
     @staticmethod
     def _get_backend_from_config(
         cfg_section: configparser.SectionProxy,
-        cfg: str
-    ) -> _Backend:
+        cfg: Union[str, os.PathLike]
+    ) -> Optional[_Backend]:
```
ZLLentz, line 880: Comment pre-dating PR: This is correct but I suspect this isn't a good interface. Other code needs to both expect that this may return None or that it may raise.

docs/source/upcoming_release_notes/362-bug_dict_invalid.rst
ZLLentz: Possibly we should also add part of my blurb from my other review comment where I explain the API differences between `SearchResult` and `InvalidResult` (to the extent that comment is correct)

ZLLentz (commented):
This makes sense and I think is well-motivated.
I'm trying to be a little cautious because it is an API break on some level.
For users with well-formed databases it will behave the same at least.

happi/client.py
tangkong: That is correct.  I could imagine enabling equality checks could be useful, but I don't think that's necessary to support.

docs/source/upcoming_release_notes/362-bug_dict_invalid.rst
tangkong: Good point. It's good to be verbose for things like this

happi/client.py
tangkong: I guess this is a matter of how severe a backend instantiation failure is to us.  I think it's sensible in this case to raise if we fail to instantiate one, since it's necessary to the operation of the client.

docs/source/upcoming_release_notes/362-bug_dict_invalid.rst
tangkong: updated in https://github.com/pcdshub/happi/pull/362/commits/7c3def5a24ab42e9fba99b1c26a6700f7946b2b1

docs/source/upcoming_release_notes/362-bug_dict_invalid.rst
```diff
@@ -0,0 +1,40 @@
...
+----------
+- Client search methods will now return an `InvalidResult` if a container (e.g.
+  `HappiItem`) cannot be created. Previously, client search methods would raise a
+  KeyError from the inciting exception. This KeyError will no longer be raised,
+  and the inciting exception will be included as a data member in the
+  InvalidResult object. This means downstream code that iterates over many happi
+  entries doesn't need special error handling on big searches unless specifically
+  the specific search item it needs is invalid.
+
+  - Similarities:
```
ZLLentz, line 14: Maybe a smidge of context here
```suggestion
  - Similarities (between SearchResult and InvalidResult):
```

ZLLentz (approved):
I'm on board. Not sure if you want to let this sit and percolate or not (it's an API break after all).

happi/client.py
tangkong: Follow up issue made #365

ZLLentz (approved):

PR #363, opened 2025-06-03
WIP/BUG: fix handling of numeric queries in search_regex

tangkong:
## Description
Handle numeric queries in `search_regex` in an intuitive way, rather than by forcing normal text searches

## Motivation and Context
#256 

There's a few edge cases I still need to iron out here, but I wanted to throw this up for visibility.  The approach I took here was to check if the provided search could be converted to a number, and if so try to perform numeric comparison.

This approach works if we have the entire document, but I don't have an analog for the mongodb backend.  In the mongo case, we construct a query to pass to mongo itself, not knowing a-priori what fields are numeric and which aren't.  This is a result of our very flexible schema.

Even after all this, we still fail to catch:
* `client.search_regex(some_field="2.0")` into `{... "some_field": "model 2.0 something"}` should probably match

## How Has This Been Tested?
Some tests have been added, but also interactively

This PR

## Pre-merge checklist
- [ ] Code works interactively
- [ ] Code contains descriptive docstrings, including context and API
- [ ] New/changed functions and methods are covered in the test suite where possible
- [ ] Test suite passes locally
- [ ] Test suite passes on GitHub Actions
- [ ] Ran ``docs/pre-release-notes.sh`` and created a pre-release documentation page
- [ ] Pre-release docs include context, functional descriptions, and contributors as appropriate

### DIFF
--- a/happi/backends/json_db.py
+++ b/happi/backends/json_db.py
@@ -261,6 +261,10 @@ class JSONBackend(_Backend):
         Find an instance or instances that matches the search criteria,
         using regular expressions.
 
+        Includes some special handling of numerics.  If the search regex can
+        be interpreted as a number, try to compare the values as numbers instead
+        of through regex matching
+
         Parameters
         ----------
         to_match : dict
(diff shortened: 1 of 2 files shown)


tangkong:
test failures related to click, addressed in #362

PR #366, opened 2025-06-25
BUG: properly handle InvalidResult in cli search method

tangkong:
## Description
  Handle invalid entries gathered in cli search method, and `HappiViewMixi`

  It's possible we actually want to adjust `Client.search` to take an argument that toggles whether or not invalid entries are returned.  Other parts of our stack likely assume everything coming from .search() is valid, and the previous change breaks this assumption.

  However, this does prevent us from using that kwarg in a container, and we'd have to be very clear about that documentation.  I've sort of painted myself into a corner here.  I think I'll opt for not changing the behavior of `Client.search` for now.  If we run into significantly more issues or get reports of this behavior breaking things, I'll recosnider.
  
  ## Motivation and Context
  Previously searching for this would error our due to the new InvalidEntry.  This would prevent the display of any other valid entries.

There are some other errors that crop up from the improper handling of malformed entries in cli commands (delete, edit, etc), but those existed prior to InvalidResult and likely deserve more scrutiny than I was hoping to devote to this small bug
  
  ## How Has This Been Tested?
  Interactively
  
  This PR  (please ignore the unrelated spam at the beginning, I think that's a me problem)
  
  Before:
  image
  
  After:
  image
  
  
  ## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/366-bug_cli_invalid_entry.rst
+++ b/docs/source/upcoming_release_notes/366-bug_cli_invalid_entry.rst
@@ -1,0 +1,22 @@
+366 bug_cli_invalid_entry
+#########################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Handle invalid entries gathered in cli search method.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tangkong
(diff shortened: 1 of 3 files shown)


tangkong:
Marking this ready for review.  Any thoughts on the quandary presented in the PR description would be appreciated (whether or not to adjust `Client.search()` to allow users to specify whether or not they want `InvalidResult` objects)

ZLLentz (approved):
I think this is super reasonable. I like how the cli helps you understand what's valid and what's not valid and making sure we only consider valid search results in the gui makes sense too.

Maybe we can talk more broadly about the related topics offline.

